PR #1, opened 2022-07-16
update submodule model

jhoetter:
new hash for submodule/model, works now with updates `docs`.


jhoetter:
test with https://github.com/code-kern-ai/refinery-submodule-model/pull/1

PR #3, opened 2022-07-21
Adds checks for demo mode

### LINKED ISSUE #7: Typo in name of mutation create_oulier_slice
https://github.com/code-kern-ai/refinery-gateway/blob/4405d31fcd030012ae02549a1f192bac9c292901/graphql_api/mutation/data_slice.py#L139

Should be called outlier instead of oulier
Has to be changed in frontend, too.

SirDegraf:
This PR introduces changes to adapt to the planned demo mode. Depending on the config-service flag for is_demo the access to gateway functionality should be restricted.

Config PR: https://github.com/code-kern-ai/refinery-config/pull/2
frontend PR: https://github.com/code-kern-ai/refinery-ui/pull/9

Also fixes https://github.com/code-kern-ai/refinery-gateway/issues/7


graphql_api/query/misc.py
```diff
@@ -30,3 +32,6 @@ def resolve_all_users_activity(self, info) -> List[UserActivityWrapper]:
 
     def resolve_is_managed(self, info) -> bool:
         return manager.check_is_managed()
+
+    def resolve_is_demo(self, info) -> bool:
+        return auth.check_is_demo(info)
```
JWittmeyer: this should be the config value because it would result in an error otherwise

controller/auth/manager.py
```diff
@@ -78,3 +81,10 @@ def check_is_admin(request: Any) -> bool:
         ):
             return True
     return False
+
+
+def check_is_demo(info: Any) -> bool:
```
JWittmeyer: doesn't return a boolean

JWittmeyer (approved):

PR #4, opened 2022-07-21
Splits app.py into modularized files

SirDegraf:
This PR outsources http endpoint definitions and middleware definitions out of app.py and creates its own structure for them so that they only are imported into app.py. 

Tested Upload for notify endpoint and all of the SDK functionality.


api/transfer.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class Notify(HTTPEndpoint):
+    async def post(self, request) -> PlainTextResponse:
+        data = await request.json()
+        file_path = data["Key"]
+
+        if len(file_path.split("/")) != 4:
+            # We need handling for lf execution notification here.
+            # ATM we have a different path of handling in util/payload_scheduler.py update_records method
```
SirDegraf, line 34: maybe this should become an issue instead of a comment, since its here for quite some while

api/transfer.py
```diff
@@ -0,0 +1,152 @@
...
+        notification.send_organization_update(project_id, "project_update", True)
+        return PlainTextResponse("OK")
+
+
+class FileExport(HTTPEndpoint):
+    def get(self, request) -> JSONResponse:
+        project_id = request.path_params["project_id"]
+        user_id = request.query_params["user_id"]
+        num_samples = request.query_params.get("num_samples")
+        auth_manager.check_project_access_from_user_id(user_id, project_id)
```
SirDegraf, line 64: These HTTP endpoints need an check for demo mode in the other PR too

JWittmeyer (approved):

PR #6, opened 2022-07-25
Fixes label upload error

SirDegraf:
While uploading with records with and without labels it could come to
the sitation where records got labeled with empty values

solves https://github.com/code-kern-ai/refinery-gateway/issues/5

sample data sets which should now work
[label-bug.json.zip](https://github.com/code-kern-ai/refinery-gateway/files/9180669/label-bug.json.zip)
[clickbait mini 2 - Labels and non-labels.json.zip](https://github.com/code-kern-ai/refinery-gateway/files/9180672/clickbait.mini.2.-.Labels.and.non-labels.json.zip)

### DIFF
--- a/controller/transfer/record_transfer_manager.py
+++ b/controller/transfer/record_transfer_manager.py
@@ -237,6 +237,9 @@ def split_record_data_and_label_data(
         label_data = {}
         for imported_key, item in data_item.items():
             if "__" in imported_key:
+                if not __check_label_has_letters(item):
+                    continue
+
                 task_name = controller.labeling_task.util.infer_labeling_task_name(
                     imported_key
                 )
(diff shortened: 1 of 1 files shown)


controller/transfer/record_transfer_manager.py
```diff
@@ -256,6 +259,12 @@ def split_record_data_and_label_data(
     return records_data, labels_data, tasks_data
 
 
+def __check_label_has_letters(label_value: str) -> bool:
+    if label_value == "" or label_value.isspace():
```
JWittmeyer: To prevent other whitespace issues i think we should stip the value (maybe for the whole column so " test " is used as "test".

if we do it one by one then with:
```suggestion
    if label_value.strip() == "":
```

SirDegraf: Its a great suggestion, but we would have to make sure that this is the behavior we want. I remember that we wanted to make as little changes to the uploaded data as possible. If we can ensure that the space is an error in all or most of the times then I would go with your solution.

JWittmeyer (approved):

PR #13, opened 2022-08-05
Adds import of exported file for knowledge terms

SirDegraf:
If a file gets imported (for knowledge terms) first the app
will try to read it in the strucutre of exported lists.
If the imported file don't have a fitting structure it gets canceled and the
original import flow will continue

Solves https://github.com/code-kern-ai/refinery/issues/61

### DIFF
--- a/controller/transfer/knowledge_base_transfer_manager.py
+++ b/controller/transfer/knowledge_base_transfer_manager.py
@@ -52,7 +52,9 @@ def import_knowledge_base_file(project_id: str, task: UploadTask) -> None:
     general.commit()
 
 
-def import_exported_file(project_id, knowledge_base_id, df):
+def import_exported_file(
+    project_id: str, knowledge_base_id: str, df: pd.DataFrame
+) -> None:
     """
     try to import the structure of export
     if the structure does not fit an exception gets raised
--- a/controller/transfer/knowledge_base_transfer_manager.py
+++ b/controller/transfer/knowledge_base_transfer_manager.py
@@ -11,6 +11,8 @@ def import_knowledge_base_file(project_id: str, task: UploadTask) -> None:
     task_manager.update_task(
         project_id, task.id, state=enums.UploadStates.PENDING.value
     )
+    general.commit()
+
     file_type = task.file_name.rsplit("_", 1)[0].rsplit(".", 1)[1]
     list_id = task.file_name.rsplit("_", 1)[1]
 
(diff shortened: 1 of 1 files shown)


controller/transfer/knowledge_base_transfer_manager.py
```diff
@@ -11,6 +11,8 @@ def import_knowledge_base_file(project_id: str, task: UploadTask) -> None:
     task_manager.update_task(
         project_id, task.id, state=enums.UploadStates.PENDING.value
     )
+    general.commit()
```
JWittmeyer, line 14: why was this added?

controller/transfer/knowledge_base_transfer_manager.py
```diff
@@ -32,13 +34,42 @@ def import_knowledge_base_file(project_id: str, task: UploadTask) -> None:
...
+            term.comment = value["comment"]
+            term.blacklisted = value["blacklisted"]
+        else:
+            knowledge_term.create(
+                project_id,
+                knowledge_base_id,
+                value["value"],
+                value["comment"],
+                value["blacklisted"],
+            )
```
JWittmeyer, line 75: i think we need a commit at the end of the function to keep it similar to the create_by_value_list. (The commit is handled in the function as well)

controller/transfer/knowledge_base_transfer_manager.py
```diff
@@ -32,13 +34,42 @@ def import_knowledge_base_file(project_id: str, task: UploadTask) -> None:
 
     if os.path.exists(download_file_name):
         os.remove(download_file_name)
-    term_list = set(df["value"].unique())
-    to_add = term_list - existing_names
 
-    knowledge_term.create_by_value_list(project_id, list_id, to_add, with_commit=True)
+    try:
+        import_exported_file(project_id, list_id, df)
+    except Exception:
```
JWittmeyer, line 40: not a fan of running into an exception only to catch it right after the function call.
Two reasons for this: 
1. Perfomance-wise exception handling is rather expensive
2. This can result in unseen issues since we catch every possible exception and don't filter for the specific one. Since we also don't print the stack trace here (because it's "expected" to fail) we wouldn't know if something goes wrong.

Let's check the structure beforehand and decide if we go with the "import exported file" or the "normal" import

SirDegraf: So I thought about different strategies and came to the conclusion that IMO running in an exception is the better way because in this scenario it is Easier to Ask for Forgiveness than Permission.

Peformance-wise it is a rather uncritical application if you think about what will be uploaded here. To check the structure beforehand is not necessary more performant. 

The Exception Handling can be done more gradually, but if there is a problem with the dataset, there will be an exception raised in the "normal" flow of import which then gets raised to the top level. 

So I would prefer to keep it like this

controller/transfer/knowledge_base_transfer_manager.py
SirDegraf: Because of a rollback in case of an exception this would be rollbacked too if not commited

controller/transfer/knowledge_base_transfer_manager.py
SirDegraf: the commit will atm be done in the calling method in line 52

controller/transfer/knowledge_base_transfer_manager.py
JWittmeyer: but inside the function there it's added with the with_commit flag so it's committed anyway, right?

SirDegraf: yes, if we trust that this will not be changed sometimes than it can be removed here

controller/transfer/knowledge_base_transfer_manager.py
JWittmeyer: Every source I could find as well as my own experience tells me that if / else is used for basically everything we know in advance. Try catch is for everything unexpected or unpredictable. Since this is also designed to fail for most import cases (the tooltip in frontend talks about a different structure for the import) this seems to be the wrong approach.

Last thought: The performance difference between a caught exception and a precondition is usually quite big. So even if it's more than a single boolean to be checked it's still noticable

SirDegraf: Just to give some sources:

https://docs.python.org/3.5/glossary.html#term-eafp

https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

There are many more if you search for "python eafp" 

In what dimension do you speak of performance? Nanoseconds? Milliseconds? Is it in the slightest sense noticeable for the user?

JWittmeyer: Just because there is the option to work with eafp doesn't mean it's always the right approach.
You are right that exceptions are handled way better in python than in some other languages. With python 3.11 there will be almost no impact on performance when no exception is being raised (https://github.com/python/cpython/issues/84403). The exact amount varies but for a singular task without a loop (like in this case) neglectable. 

For when to use eafp vs lbyl I bet we will both find cases to support our approach, e.g.:
https://realpython.com/python-lbyl-vs-eafp/#eafp-vs-lbyl-with-examples

> Use LBYL for -> Operations that are likely to fail 

Since this wont impact the workflow in any meaningful way this can be either ignored or changed. We should however at some point agree on basic guidelines regarding which approach to choose depending on the situation (add something in the notion page) so it's not a discussion point.

JWittmeyer (approved):

PR #21, opened 2022-08-08
Dependabot pip numpy 1.22.0

JWittmeyer:


JWittmeyer (commented):
bump further to 1.23.1

JWittmeyer:
bumped back down to 1.22.04 since other libs don't work with a higher version of numpy

FelixKirsch (approved):

PR #25, opened 2022-08-11
Adds qutation escaping

JWittmeyer:


service/search/search.py
```diff
@@ -164,6 +166,21 @@ def resolve_extended_search(
     return extended_search
 
 
+def __ensure_text(filter_data: List[Dict[str, Any]]) -> None:
```
FelixKirsch:
```suggestion
def __ensure_text(filter_data: List[str | Dict[str, Any]]) -> None:
```
Change typing, to clarify that the list can contain strings or dictionaries

FelixKirsch (approved):

PR #32, opened 2022-08-25
Integration for the model provider service

FelixKirsch:
works with:
- https://github.com/code-kern-ai/refinery-embedder/pull/9
- https://github.com/code-kern-ai/refinery-config/pull/6
- https://github.com/code-kern-ai/refinery-ui/pull/44
- https://github.com/code-kern-ai/refinery-zero-shot/pull/6
- https://github.com/code-kern-ai/dev-setup/pull/15


controller/model_provider/manager.py
```diff
@@ -0,0 +1,26 @@
...
+BASE_URI = os.getenv("MODEL_PROVIDER")
+
+
+def get_model_provider_info() -> List[ModelProviderInfoResult]:
+    url = f"{BASE_URI}/info"
+    return service_requests.get_call_or_raise(url)
+
+
+def model_provider_download_model(
+    project_id: str, model_name: str, revision: str
```
JWittmeyer: why do we need a project_id and revision to download a model?

FelixKirsch: I added it for the websocket notifications, is there a better way to use websockets without the project id?

lumburovskalina: The revision is not need, will remove it, but the project_id is needed for the websocket

FelixKirsch: Regarding the revision, at the beginning I implemented the model-provider service so that it is also capable of managing multiple revisions of the same model, as this is not the plan anymore, we could remove this functionality in the service itself and the gateway

JWittmeyer: project_id in notifications is somewhat problematic since models are globally available and with this only the specific project would be notified. If person A downloads in project A a model and person B somewhat simultaneously tries to download the same model in project B this will result in some issues. 

I'd go without the project_id and a global subscription in the frontend. However, this would only solve the issue for multiple projects within the same org. For multiple orgs multiple notifications need to be sent (one for each org). If there are questions regarding the how please ask :)

graphql_api/types.py
```diff
@@ -682,3 +682,13 @@ class ZeroShotNRecords(graphene.ObjectType):
...
     duration = graphene.Float()
     records = graphene.List(ZeroShotNRecords)
+
+
+class ModelProviderInfoResult(graphene.ObjectType):
+    name = graphene.String()
+    revision = graphene.String()
+    link = graphene.String()
+    date = graphene.Float()
+    size = graphene.Int()
```
JWittmeyer: i think this is the byte size of the model. For other size objects, we decided to use Float to ensure the value fits.(see byte_size field)

graphql_api/types.py
```diff
@@ -682,3 +682,13 @@ class ZeroShotNRecords(graphene.ObjectType):
 class ZeroShotNRecordsWrapper(graphene.ObjectType):
     duration = graphene.Float()
     records = graphene.List(ZeroShotNRecords)
+
+
+class ModelProviderInfoResult(graphene.ObjectType):
+    name = graphene.String()
+    revision = graphene.String()
+    link = graphene.String()
+    date = graphene.Float()
```
JWittmeyer: there is a graphene.Date object, why wasn't this used?

util/service_requests.py
```diff
@@ -22,8 +22,8 @@ def get_call_or_raise(url: str, params: Dict = {}) -> Any:
         raise GraphQLError(response.text)
 
 
-def delete_call_or_raise(url: str) -> int:
-    response = requests.delete(url=url)
+def delete_call_or_raise(url: str, params: Dict = {}) -> int:
```
JWittmeyer: Dont use a Dict as a default value (see [notion](https://www.notion.so/kern-ai/Coding-Guide-Python-f54c6201aa22471caaff38d765c0d64e#8ea49ab353b54cb3960cadafb4307389) for reference)

controller/model_provider/manager.py
JWittmeyer: @FelixKirsch please do 🙂

graphql_api/types.py
FelixKirsch: changed

graphql_api/types.py
FelixKirsch: changed

util/service_requests.py
FelixKirsch: changed

controller/model_provider/manager.py
FelixKirsch: changed in gateway, some changes in the model-provider are still needed

JWittmeyer:
I'm guessing this is for the model provider
image

During the download zero shot models are listed as NOT ZS (since the readme isn't there yet I guess).
Since in the future only an icon is displayed for the ZS models nothing to change atm. 
In the future we might be able to detect that before everything is fully downloaded

- [x] issue created in model provider

JWittmeyer (approved):

JWittmeyer (approved):

PR #39, opened 2022-09-11
External information sources

jhoetter:


jhoetter:
Run this together with:
- [UI PR](https://github.com/code-kern-ai/refinery-ui/pull/51)
- [Proxy PR](https://github.com/code-kern-ai/refinery-gateway-proxy/pull/6)
- [Submodule PR](https://github.com/code-kern-ai/refinery-submodule-model/pull/7)

You can add a model callback using the Python SDK in branch `external-information-sources` ([PR](https://github.com/code-kern-ai/refinery-python-sdk/pull/22)) as follows:
```python
from refinery import Client

user_name = "your-mail" # this is the email you log in with
password = "your-pw"
project_id = "your-clickbait-sample-project-id" # can be found in the URL of the web application

client = Client(user_name, password, project_id, uri="http://localhost:4455")

from sklearn.linear_model import LogisticRegression
clf = LogisticRegression()

from refinery.adapter.sklearn import build_classification_dataset
data = build_classification_dataset(client, "headline", "__clickbait", "distilbert-base-uncased", num_train=50)
clf.fit(data["train"]["inputs"], data["train"]["labels"])

from refinery.callbacks.sklearn import SklearnCallback
callback = SklearnCallback(
    client, 
    clf,
    "clickbait", 
)

callback.run(data["train"]["inputs"], data["train"]["index"])
callback.run(data["test"]["inputs"], data["test"]["index"])
```

api/transfer.py
```diff
@@ -98,6 +99,29 @@ async def post(self, request) -> JSONResponse:
...
+    async def post(self, request) -> JSONResponse:
+        project_id = request.path_params["project_id"]
+        request_body = await request.json()
+        user_id = request_body["user_id"]
+        auth_manager.check_project_access_from_user_id(user_id, project_id)
+        associations = request_body["associations"]
+        indices = request_body["indices"]
+        name = request_body["name"]
+        label_task_name = request_body["label_task_name"]
+        source_type = request_body["source_type"]
```
JWittmeyer: personal preference: why are these stored in variables when the dict key is = to the variable name & only used once for a function call?

controller/information_source/manager.py
```diff
@@ -8,12 +8,16 @@ def get_information_source(project_id: str, source_id: str) -> InformationSource
...
+    return information_source.get_by_name(project_id, name)
+
+
 def get_all_information_sources(project_id: str) -> List[InformationSource]:
     return information_source.get_all(project_id)
 
 
-def get_overview_data(project_id: str) -> str:
-    return information_source.get_overview_data(project_id)
+def get_overview_data(project_id: str, operator: Optional[str] = "!=") -> str:
```
JWittmeyer: optional is only for = None 
[notion link](https://www.notion.so/kern-ai/Coding-Guide-Python-f54c6201aa22471caaff38d765c0d64e#447b4b18e8b94725b8ad0c767445fcb3)

JWittmeyer: also, im not the biggest fan of operator as variable since it's pretty undescriptive what it's for.

What do you think about `,for_model_callback: bool = false)->str` and in the information source function something like
``` python
type_selection = " != 'MODEL_CALLBACK'"
if for_model_callback:
     type_selection = " = 'MODEL_CALLBACK'"
```

same for update_is_selected_for_project

controller/transfer/association_transfer_manager.py
```diff
@@ -0,0 +1,111 @@
...
+        elif source_type == "heuristic":
+            information_source = information_source_manager.create_information_source(
+                project_id,
+                user_id,
+                labeling_task.id,
+                model_name,
+                "",
+                "This is a heuristic",
+                enums.LabelSource.INFORMATION_SOURCE.value,
+            )
```
JWittmeyer: pretty similar maybe:
```suggestion
        
        if source_type == "model_callback":
            description = "This is a model callback"
            type = enums.LabelSource.MODEL_CALLBACK.value
        elif source_type == "heuristic":
            description = "This is a heuristic"
            type = enums.LabelSource.INFORMATION_SOURCE.value
            
        information_source = information_source_manager.create_information_source(
                project_id,
                user_id,
                labeling_task.id,
                model_name,
                "",
                description,
                type,
            )
```

controller/transfer/association_transfer_manager.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    record_ids = [record.id for record in list(records.values())]
+
+    record_label_associations = []
+    for record_id, association in zip(record_ids, associations):
+        label_name, confidence = association
+
+        label = labeling_task_label.get_by_name(
+            project_id, labeling_task.id, label_name
+        )
```
JWittmeyer: this will run a new query every loop, i'd go with get all, dict comprehension and grabbing the data then.

This would also open the door for list comprehension for the record_label_associations which will struggle with the append on bigger lists.

Also depending on the size it might make sense to batch this a bit sind an insert with > 10k records will run into issues. maybe max 1k records per batch

controller/transfer/association_transfer_manager.py
```diff
@@ -0,0 +1,111 @@
...
+    general.commit()
+
+    try:
+        weak_supervision.calculate_stats_after_source_run(
+            project_id, information_source.id, user_id
+        )
+        notification.send_organization_update(
+            project_id, f"information_source_statistics:{information_source.id}"
+        )
+        general.commit()
```
JWittmeyer: unnecessary

controller/weak_supervision/weak_supervision_service.py
```diff
@@ -34,7 +36,10 @@ def calculate_quality_after_labeling(project_id: str, labeling_task_id: str, use
     return service_requests.post_call_or_raise(url, data)
 
 
-def calculate_stats_after_source_run(project_id: str, source_id: str, user_id: str) -> Any:
+@debounce(2)
```
JWittmeyer: if i'm not mistaken this was intentionally not debounced to have an "on demand" execution -> debounce will start a thread and not wait for the return. Is this wanted?

graphql_api/mutation/information_source.py
```diff
@@ -91,6 +91,23 @@ def mutate(self, info, project_id: str, value: bool):
...
+        value = graphene.Boolean(required=True)
+
+    ok = graphene.Boolean()
+
+    def mutate(self, info, project_id: str, value: bool):
+        auth.check_demo_access(info)
+        auth.check_project_access(info, project_id)
+        manager.set_all_model_callbacks_selected(project_id, value)
+        notification.send_organization_update(
+            project_id, f"information_source_updated:all"
```
JWittmeyer: is this correct? (information_source_update) this will trigger an update on the heuristics page as well

controller/transfer/association_transfer_manager.py
```diff
@@ -0,0 +1,111 @@
...
+                source_id=information_source.id,
+                return_type=enums.InformationSourceReturnType.RETURN.value,
+                confidence=confidence,
+                created_by=user_id,
+            )
+        )
+
+    record_label_association.delete_by_source_id_and_record_ids(
+        project_id, information_source.id, record_ids
+    )
```
JWittmeyer, line 93: is this wanted? only the provided record ids are removed so a newer import with only 10 record ids would still look like it has a lot of records corresponding to it?

api/transfer.py
jhoetter: true, changed that

controller/information_source/manager.py
jhoetter: makes sense, changed that

controller/transfer/association_transfer_manager.py
jhoetter: the batch is limited to 32 files per request. I added the change re the label query calling every loop iteration into one before the loop :)

controller/transfer/association_transfer_manager.py
jhoetter: removed it

controller/weak_supervision/weak_supervision_service.py
jhoetter: The data is sent from the SDK in smaller batches, so I don't want to recompute the stats every time (this would again result in an overload). 2 seconds for debounce should be ok I guess. An alternative would be to add another function calling this, which receives the debounce, and only having the callback execute this very function.

jhoetter: nvm I changed it so that only the callbacks have a debounce

graphql_api/mutation/information_source.py
jhoetter: No, only meant as an update for the model callbacks. i commented it out

controller/transfer/association_transfer_manager.py
jhoetter: again, this is a result of the batching mechanism

JWittmeyer (approved):

PR #44, opened 2022-09-15
upload per json

jhoetter:
Requires [proxy](https://github.com/code-kern-ai/refinery-gateway-proxy/pull/8) and [SDK](https://github.com/code-kern-ai/refinery-python-sdk/pull/23) PR.

Try out in the SDK via:

```python
from refinery import Client

user_name = "your-username" # this is the email you log in with
password = "your-pw"
project_id = "clickbait-project-sample-id" # can be found in the URL of the web application

client = Client(user_name, password, project_id, uri="http://localhost:4455")

data = [{
    "running_id": 99998,
    "headline": "I uploaded this"
},{
    "running_id": 99999,
    "headline": "I also uploaded this one!"
}]

client.post_records(data)
```

### DIFF
--- a/controller/transfer/record_transfer_manager.py
+++ b/controller/transfer/record_transfer_manager.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, Any, Tuple, List
+from typing import Dict, Any, Optional, Tuple, List
 
 import pandas as pd
 
@@ -17,7 +17,7 @@ from submodules.model.business_objects import (
     record_label_association,
 )
 from controller.user import manager as user_manager
-from controller.upload_task import manager as task_manager
+from controller.upload_task import manager as upload_task_manager
 from controller.tokenization import manager as token_manager
 from util import doc_ock
 from submodules.s3 import controller as s3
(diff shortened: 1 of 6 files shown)


controller/transfer/record_transfer_manager.py
```diff
@@ -90,25 +69,80 @@ def import_file(project_id: str, task: UploadTask) -> None:
...
+                project_id, upload_task.id, progress=progress
+            )
+
+
+def import_record_json(
+    project_id: str, user_id: str, record_data: Dict[str, Any]
+) -> None:
+    import_records_and_rlas(project_id, user_id, record_data)
+    token_manager.request_tokenize_project(project_id, user_id)
+    token_manager.create_rats_entries(project_id, user_id)
```
JWittmeyer: shouldn't be nessecary since the tokenizer does it after the doc creation anyway and cancels the job if doc creation is running

jhoetter: When I did not add it, I had a loading spinner in the UI for this record and received the error in the gateway that there was not docbin entry for said record.

JWittmeyer:
small project (100 r) with a small addition (2 r) doesn'T seem to get tokenized (labeling page doesn't work & tables only holds the initial 100 records not the 2 additional (table record_tokenized) The record_tokenization_task also only holds the tasks for the initial 100. Guessing maybe a race condition.
image

Further proof, the second time I added the records the tokenization task is created but with only the 2 old records but 4 rats
I'm guessing we only need to move the general.commit() before the requests 

- [ ] resolved

JWittmeyer:
Project record data doesn't update (header not e.g. data browser):
image

Can be solved by sending a websocket notification. Up until now we listen to messages `project_updated` & `file_upload`.

A solution would be to add a websocket notification & change the frontend listener (though that can be done in a different branch :) ) 

- [ ] resolved

JWittmeyer:
data isn't checked if it should be accepted.
Testcase 100.000 records via json upload shouldn't be possible.
It still tries to upload everything -- and after a few batches fails with
(Sdk)
image
(tokenizer)
image
image

I'm guessing since the data is batched and the requests to tokenized a project pile up this should also be fixed:
Possible solution: In the sdk send a "is last package" with the request so the gateway knows when to start tokenization

This will also solve some weird looking percentage values
image


- [ ] resolved checks before transfering
- [ ] resolved tokenized only once

JWittmeyer:
Uploading records with a different column set over the UI test for existing columns
image

the sdk just accepts them and runs no checks (they are in the method `convert_to_record_dict`)

- [ ] resolved

JWittmeyer:
maybe to not rebuild everything we could "collect" the batched record data in the gateway, create a task entry and add the file to s3 storage to run the usual route (e.g. by creating a uuid for the upload request and the "last batch" flag)

- [x] discussed
- [ ] ignored
- [x] resolved

JWittmeyer (approved):

jhoetter:
@JWittmeyer and I agreed upon changing the logic to deal with this as a file import

PR #46, opened 2022-09-20
avoid setting overlapping extraction labels

jhoetter:
together with [frontend](https://github.com/code-kern-ai/refinery-ui/pull/57) and [submodule](https://github.com/code-kern-ai/refinery-submodule-model/pull/11)

### DIFF
--- a/controller/record_label_association/manager.py
+++ b/controller/record_label_association/manager.py
@@ -111,7 +111,31 @@ def create_manual_extraction_label(
     if label_item is None:
         return None
 
-    tokens = record_label_association.create_token_objects(
+    existing_tokens = record_label_association.get_manual_tokens_by_record_id(record_id)
+
+    tokens = []
+    curr_start = None
+    curr_end = None
+    for existing_token in existing_tokens:
+        if existing_token.is_beginning_token:
+            if curr_start is not None:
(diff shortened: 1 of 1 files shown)


controller/record_label_association/manager.py
```diff
@@ -111,7 +111,31 @@ def create_manual_extraction_label(
...
+    curr_start = None
+    curr_end = None
+    for existing_token in existing_tokens:
+        if existing_token.is_beginning_token:
+            if curr_start is not None:
+                tokens.append([curr_start, curr_end])
+            curr_start = existing_token.token_index
+        curr_end = existing_token.token_index
+    if curr_start is not None:
+        tokens.append([curr_start, curr_end])
```
FelixKirsch, line 126: if I do not miss anything, this only works if the tokens are sorted by token_index, is this guaranteed by the query?

jhoetter: generally, they are inserted almost instantly, so i would assume that in 99.99...% of the cases, they are in order. But you are right, if two users label simultaneously on reference data (not talking about crowd labelers, this is not handled as manual labeling), this might not be guaranteed. Will treat this as an issue, as we need to modify the database model for that (i.e. adding created_by and created_at

PR #55, opened 2022-09-28
Release v1.3.4

lumburovskalina:
- https://github.com/code-kern-ai/refinery-ui/pull/66
- https://github.com/code-kern-ai/refinery-gateway/pull/55
- https://github.com/code-kern-ai/refinery-submodule-model/pull/17

### DIFF
--- a/controller/attribute/manager.py
+++ b/controller/attribute/manager.py
@@ -87,7 +87,8 @@ def update_attribute(
     name: str,
     source_code: str,
 ) -> None:
-    attribute.update(
+
+    attribute_item: Attribute = attribute.update(
         project_id,
         attribute_id,
         data_type,
@@ -96,12 +97,10 @@ def update_attribute(
         source_code,
         with_commit=True,
     )
-    if attribute.get(project_id, attribute_id).state in [
-        AttributeState.UPLOADED.value,
-        AttributeState.AUTOMATICALLY_CREATED.value,
-        AttributeState.USABLE.value,
-    ]:
-        notification.send_organization_update(project_id, "attributes_updated")
+
+    notification.send_organization_update(
+        project_id=project_id, message=f"calculate_attribute:updated:{str(attribute_item.id)}"
+    )
 
 
 def delete_attribute(project_id: str, attribute_id: str) -> None:
--- a/controller/payload/manager.py
+++ b/controller/payload/manager.py
@@ -1,8 +1,7 @@
-from typing import Any, Optional
-
+from typing import List, Optional, Tuple
 from controller.payload import payload_scheduler
-from submodules.model import InformationSourcePayload
-from submodules.model.business_objects import payload
+from submodules.model import InformationSourcePayload, enums
+from submodules.model.business_objects import information_source, payload
 
 
 def get_payload(project_id: str, payload_id: str) -> InformationSourcePayload:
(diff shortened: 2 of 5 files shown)


controller/attribute/manager.py
```diff
@@ -96,12 +97,10 @@ def update_attribute(
...
-    if attribute.get(project_id, attribute_id).state in [
-        AttributeState.UPLOADED.value,
-        AttributeState.AUTOMATICALLY_CREATED.value,
-        AttributeState.USABLE.value,
-    ]:
-        notification.send_organization_update(project_id, "attributes_updated")
+
+    notification.send_organization_update(
+        project_id=project_id, message=f"calculate_attribute:updated:{str(attribute_item.id)}"
+    )
```
JWittmeyer, line 103: i think we will run into some merge issues here (since I had to change the websocket notification a bit as well.
-> it's not wrong, just a reminder for us so we don't forget it :)

controller/payload/manager.py
```diff
@@ -39,3 +42,43 @@ def update_payload_status(
...
+        ],
+        container_logs=container_logs,
+        code_has_errors=code_has_errors,
+    )
+
+
+def fill_missing_record_ids(sample_records: List[str], calculated_labels) -> List[str]:
+    for record_item in sample_records:
+        record_id = record_item[0]
+        if record_id not in calculated_labels.keys():
```
JWittmeyer: since caluclated labels is a dict a lookup without .keys would be better
https://stackoverflow.com/a/4731079/19801189

controller/payload/manager.py
```diff
@@ -39,3 +42,43 @@ def update_payload_status(
...
+                calculated_labels=calculated_labels[record_item[0]],
+            )
+            for record_item in sample_records
+        ],
+        container_logs=container_logs,
+        code_has_errors=code_has_errors,
+    )
+
+
+def fill_missing_record_ids(sample_records: List[str], calculated_labels) -> List[str]:
```
JWittmeyer: typing for calcualted_labels missing

controller/payload/payload_scheduler.py
```diff
@@ -694,3 +697,99 @@ def add_information_source_statistics_exclusion(
...
+
+
+def prepare_sample_records_doc_bin(
+    project_id: str, information_source_id: str
+) -> Tuple[str, List[str]]:
+    sample_records = record.get_attribute_calculation_sample_records(project_id)
+
+    sample_records_doc_bin = get_doc_bin_table_to_json(
+        project_id=project_id,
+        missing_columns="",
```
JWittmeyer: if i'm not toally mistaken this will result in some errors when we access the data of a collumn outside the docbins (e.g. running_id) since the missing_columns aren't provided

controller/payload/payload_scheduler.py
```diff
@@ -694,3 +697,99 @@ def add_information_source_statistics_exclusion(
...
+    s3.put_object(
+        org_id,
+        project_id + "/" + prefixed_doc_bin,
+        sample_records_doc_bin,
+    )
+
+    return prefixed_doc_bin, sample_records
+
+
+def run_labeling_function_exec_env(
```
JWittmeyer, line 725: this feels like a lot of duplicated code.
can we reuse stuff from run_container ? or create some methods that can be used in both?

lumburovskalina: Closing this for now, but we have to change it

JWittmeyer (approved):

PR #59, opened 2022-09-30
adds comment logic

JWittmeyer:
- https://github.com/code-kern-ai/refinery-ui/pull/69
- https://github.com/code-kern-ai/refinery-submodule-model/pull/15
- https://github.com/code-kern-ai/refinery-gateway/pull/59

### DIFF
--- a/controller/transfer/project_transfer_manager.py
+++ b/controller/transfer/project_transfer_manager.py
@@ -23,6 +23,7 @@ from submodules.model.business_objects import (
     user,
     knowledge_term,
     weak_supervision,
+    comments as comment,
 )
 from submodules.model.enums import NotificationType
 from controller.labeling_access_link import manager as link_manager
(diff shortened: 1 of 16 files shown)


controller/transfer/project_transfer_manager.py
```diff
@@ -129,14 +130,14 @@ def import_sample_project(
...
-            project_item.id, "project_update", is_global=True
+            project_item.id, f"project_update:{str(project_item.id)}", is_global=True
         )
         return project_item
 
 
 def import_file(
     project_id: str,
-    user_id: str,
+    import_user_id: str,
```
FelixKirsch, line 140: user_id has been used in lines 494 and 827 and needs to be replaced there with import_user_id.
Don't throw errors currently, because user_id is defined somewhere else in the method, but this is not intended.

graphql_api/mutation/comment.py
```diff
@@ -6,11 +6,15 @@
 from controller.comment import manager
 
 
+def send_notification():
+    pass
```
FelixKirsch: is this a preparation for further development?

JWittmeyer: nope, deleted :)

controller/transfer/project_transfer_manager.py
JWittmeyer: That's why i changed it but missed two spots. good catch :)

FelixKirsch (approved):

PR #64, opened 2022-10-13
Fixes record upload logic for record upload

### LINKED ISSUE #63: Importing records into an existing project fails
Steps to reproduce:
* Create a sample project, e.g. "Clickbait - initial"
* Label a record
* Download the records as export.json
* Upload the records
* Observe record import fails

The following is seen in the logs for the refinery-gateway docker container:

```
  File "/app/./api/transfer.py", line 53, in post
    init_file_import(task, project_id, is_global_update)
  File "/app/./api/transfer.py", line 122, in init_file_import
    upload_manager.import_records(project_id, task)
  File "/app/./controller/transfer/manager.py", line 47, in import_records
    import_file(project_id, task)
  File "/app/./controller/transfer/record_transfer_manager.py", line 86, in import_file
    ) = split_record_data_and_label_data(chunk)
  File "/app/./controller/transfer/record_transfer_manager.py", line 241, in split_record_data_and_label_data
    item.strip() == ""
AttributeError: 'list' object has no attribute 'strip'

```

My use case here is that I wanted to re-import the labelling I have done in a similar project into a new project with a schema with an additional field.  It failed in a similar way (float has no attribute 'strip')  I am unsure if this use case is supported, but it seems like roundtripping the records should work if a record has a primary key defined.

### LINKED ISSUE #105: [BUG] - Primary key update logic seems faulty
**Describe the bug**
In the sample project clickbait data, I wanted to upload new dummy data with the following logic:
```python
import pandas as pd

data = pd.DataFrame([{
    "running_id": idx,
    "headline": "I uploaded this"
} for idx in range(1_000)])
```

Now, when I have the `running_id` as the primary key, what I would **expect** is that my existing records are overwritten. What I get, however, is that 500 records are written new to the database, and 500 seem to be overwritten.

**To Reproduce**
1. Jump into the clickbait sample project
2. Either upload a `.json` file containing the above generated data, or use the SDK (`client.post_df(data)`)

**Expected behavior**
See above.

**Screenshots**
Bildschirmfoto 2022-09-21 um 16 53 21

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser: Chrome
 - Version: 1.2.0

**Additional context**
\-

SirDegraf:
Fixes record upload logic for second and following chunks by making the primary keys available in later chunks too

closes [[BUG] - Primary key update logic seems faulty #105](https://github.com/code-kern-ai/refinery/issues/105)

Steps to solve:

1. analyze and understand error (1-2 hours)
2. If easy to fix, continue / if hard to fix, reevaluate time estimations (15 min)
3. Create Branch and PR, setup implementation (15 min)
4. Implement fix (1 hour)
5. Test fixes (1 hour)
6. Review PR (1 hour)
7. Merge changes and close issue (15 min)

-> 6-7 hour estimation + 1 hour of buffer, so roughly a day in total

Datasets to reproduce and test: [test-sets.zip](https://github.com/code-kern-ai/refinery-gateway/files/9775463/test-sets.zip)

The bug occurs on the 500th record and following ones because primary keys got only read for the first chunk. Made the primary keys available for following chunks.

Further finding: If a record gets uploaded with a new attribute (which is not yet present in the project) it gets created and all of the other records get an empty attribute field added . Don't know yet if this is an issue at all or is wanted, so @JWittmeyer maybe we have to talk about this, because if this is not wanted it could maybe solved in this PR, too! Can be reproduced by creating a sample Clickbait project and then upload the dataset "clickbait_abc_extension_1.json" out of the referenced [test-sets.zip](https://github.com/code-kern-ai/refinery-gateway/files/9775463/test-sets.zip)/1 into the clickbait project.

### DIFF
--- a/controller/transfer/record_transfer_manager.py
+++ b/controller/transfer/record_transfer_manager.py
@@ -66,7 +66,7 @@ def import_records_and_rlas(
         
         if idx == 0:
             create_attributes_and_get_text_attributes(project_id, records_data)
-        primary_keys = attribute.get_primary_keys(project_id)
+            primary_keys = attribute.get_primary_keys(project_id)
 
         import_labeling_tasks_and_labels_pipeline(
             project_id=project_id, tasks_data=tasks_data
--- a/controller/transfer/record_transfer_manager.py
+++ b/controller/transfer/record_transfer_manager.py
@@ -63,11 +63,10 @@ def import_records_and_rlas(
             labels_data,
             tasks_data,
         ) = split_record_data_and_label_data(chunk)
-
-        primary_keys = None
+        
         if idx == 0:
             create_attributes_and_get_text_attributes(project_id, records_data)
-            primary_keys = attribute.get_primary_keys(project_id)
+        primary_keys = attribute.get_primary_keys(project_id)
 
         import_labeling_tasks_and_labels_pipeline(
             project_id=project_id, tasks_data=tasks_data


controller/transfer/record_transfer_manager.py
```diff
@@ -63,11 +63,10 @@ def import_records_and_rlas(
             labels_data,
             tasks_data,
         ) = split_record_data_and_label_data(chunk)
-
-        primary_keys = None
+        
         if idx == 0:
             create_attributes_and_get_text_attributes(project_id, records_data)
-            primary_keys = attribute.get_primary_keys(project_id)
+        primary_keys = attribute.get_primary_keys(project_id)
```
JWittmeyer: they shouldn't change within the loop so only requesting them once is valid imo. The initialization to None is faulty (or needs to be done outside the loop)

SirDegraf: You are right, this is about primary keys and they should not change during the transaction even if the user triggers a change during the execution by the frontend. Will change this back!

JWittmeyer (approved):

PR #84, opened 2022-11-24
Implementing inference api

FelixKirsch:
see:
- https://github.com/code-kern-ai/refinery-weak-supervisor/pull/12
- https://github.com/code-kern-ai/refinery-commercial-proxy/pull/1
- https://github.com/code-kern-ai/refinery-ml-exec-env/pull/15
- https://github.com/code-kern-ai/refinery-tokenizer/pull/13
- https://github.com/code-kern-ai/refinery-gateway/pull/84
- https://github.com/code-kern-ai/refinery-embedder/pull/21
- https://github.com/code-kern-ai/refinery-inference-exec-env


controller/information_source/manager.py
```diff
@@ -120,9 +123,24 @@ def update_information_source(
 
 
 def delete_information_source(project_id: str, source_id: str) -> None:
+    if information_source.get(
+        project_id, source_id
+    ).type == enums.InformationSourceType.ACTIVE_LEARNING.value and config_service.get_config_value(
```
JWittmeyer: to prevent edge cases i would get the is from the database first and only continue if != None

```suggestion
    is_item = information_source.get(project_id, source_id)
    if not is_item:
         #some error
         return
    if is_item.type == enums.InformationSourceType.ACTIVE_LEARNING.value and config_service.get_config_value(
```

controller/payload/payload_scheduler.py
```diff
@@ -60,6 +60,8 @@
 lf_exec_env_image = os.getenv("LF_EXEC_ENV_IMAGE")
 ml_exec_env_image = os.getenv("ML_EXEC_ENV_IMAGE")
 exec_env_network = os.getenv("LF_NETWORK")
+if config_service.get_config_value("is_managed"):
```
JWittmeyer: as far as i know we usually don't run conditions etc in the .py file directly. 
Since the config service also needs to be present for this to work (which it always should but who know :D ) I would go maybe with a slightly different approach. Singleton...ish:

```
inference_dir = "NOT_INITIALZED"

def get_finerence_dir():
   if inference_dir == "NOT_INITIALZED"
       if config_service.get_config_value("is_managed"):
           inference_dir = os.getenv("INFERENCE_DIR")
       else:
           inference_dir = None
   return inference_dir
```

and wherever we need the value we can use the function. This would then only collect the info if/when necessary when possible other startup issues are already handled.

controller/payload/payload_scheduler.py
```diff
@@ -183,6 +185,7 @@ def prepare_input_data_for_payload(
             training_record_ids = get_exclusion_record_ids(information_source_id)
             input_data = json.dumps(
                 {
+                    "information_source_id": str(information_source_item.id),
```
JWittmeyer: why not `information_source_id` from create_payload? though i don't know if it's already a string however the typing suggest yes :D

start
```diff
@@ -19,12 +19,19 @@ else
...
-echo -ne '\t [done]\n'
+INFERENCE_DIR=${PWD%/*}/dev-setup/inference/
+if [ ! -d "$_DIR" ] 
+then
+    INFERENCE_DIR=${PWD%/*/*}/dev-setup/inference/
+    if [ ! -d "$INFERENCE_DIR" ] 
+    then
+        # to include volume for local development, use the dev-setup model data folder:  
+        # alternative use manual logic with
+        # -v /path/to/dev-setup/model-data:/models \
```
JWittmeyer: copy comment

JWittmeyer (approved):

controller/information_source/manager.py
FelixKirsch: changed to the proposed solution

controller/payload/payload_scheduler.py
FelixKirsch: This works, if we assume that the config for "is_managed" does not change during runtime. Is that fine or should we also handle this edge case?

start
FelixKirsch: changed

controller/payload/payload_scheduler.py
FelixKirsch: I don't know :D changed it

controller/payload/payload_scheduler.py
JWittmeyer: changing at runtime isn't supposed to happen atm (all our guides to change include a restart of gateway & config service).
That said we could also go without the global variable and just use the function since the overhead for getting the config value & os variable should be negligible:
```python
# inference_dir = "NOT_INITIALZED"

def get_finerence_dir() -> str:
   if config_service.get_config_value("is_managed"):
       return os.getenv("INFERENCE_DIR")
   return None

```

FelixKirsch: changed

PR #96, opened 2022-12-19
Refactored tokenization service and and calls

SirDegraf:
Tokenizer: https://github.com/code-kern-ai/refinery-tokenizer/pull/17
Submodule Model: https://github.com/code-kern-ai/refinery-submodule-model/pull/30
Gateway: https://github.com/code-kern-ai/refinery-gateway/pull/96


controller/tokenization/tokenization_service.py
```diff
@@ -25,6 +25,18 @@ def request_tokenize_project(project_id: str, user_id: str) -> None:
...
 
 
+def request_tokenize_calculated_attribute(
+    project_id: str, user_id: str, attribute_name: str
+) -> None:
+    url = f"{BASE_URI}/tokenize_calculated_attribute"
+    data = {
+        "project_id": str(project_id),
+        "user_id": str(user_id),
+        "attribute_name": str(attribute_name),
```
JWittmeyer: is there a specific reason we use the name and not the id?

SirDegraf: Changed to ID

controller/attribute/manager.py
```diff
@@ -251,17 +252,28 @@ def __calculate_user_attribute_all_records(
...
+        try:
+            request_tokenize_calculated_attribute(
+                project_id, user_id, attribute_item.id
+            )
+        except:
+            record.delete_user_created_attribute(
+                project_id=project_id,
+                attribute_id=attribute_id,
+                with_commit=True,
+            )
```
JWittmeyer, line 269: not sure if we should simply delete the attribute if the tokenization fails maybe set the attribute to error?

JWittmeyer (approved):

PR #97, opened 2023-01-17
create refinery project from workflow store

jhoetter:
adds the following mutation:
```gql
  mutation createProjectByWorkflowStore($name: String!, $description: String!, $tokenizer: String!, $storeId: String!) {
    createProjectByWorkflowStore(
      name: $name,
      description: $description,
      tokenizer: $tokenizer,
      storeId: $storeId
    ) {
      project {
        id
      }
    }
  } 
```

which allows refinery to read the database of workflow and create a new project from the given store.

To test this, please make sure that you run the [workflow-setup](https://github.com/code-kern-ai/workflow-setup), and in the workflow ui, add a store (e.g. a google spreadsheet, or your gmail account).


graphql_api/mutation/project.py
```diff
@@ -36,6 +48,58 @@ def mutate(self, info, name: str, description: Optional[str] = None):
...
+            organization.id, name, description, user.id
+        )
+        project_manager.update_project(project_id=project.id, tokenizer=tokenizer)
+        engine = create_engine(workflow_postgres_url)
+
+        Session = sessionmaker(engine)
+        with Session() as session:
+            results = session.execute(
+                f"SELECT record FROM store_entry WHERE store_id = '{storeId}'"
+            ).all()
```
JWittmeyer: not sure regarding the approach here. This will create an engine every single time which shouldn't be done more than once per database URL -> [doc](https://docs.sqlalchemy.org/en/13/core/connections.html#basic-usage) .

A singletone would do the trick imo :) 


also with this we now have database knowledge here from the workflow app which links them quite closely. 
I think we could do it similar to uploading stuff in general => create a task and upload to minio (from workflow) which then triggers the usual routine.

jhoetter: we had the discussion with minio, but workflow is meant to be coupled super closely to refinery. So it makes perfect sense to use the direct db connection, as workflow is the data-sender for refinery.

Not an expert with the sessionmaker, but a singleton I can build :)

jhoetter: done.

JWittmeyer (approved):

PR #99, opened 2023-01-18
labeling suite backend changes

JWittmeyer:
- https://github.com/code-kern-ai/refinery-ui/pull/100
- https://github.com/code-kern-ai/refinery-gateway/pull/99

general idea is to change the backend as little as possible so only absolutely necessary changes will be introduces


controller/record_label_association/manager.py
```diff
@@ -184,35 +184,10 @@ def create_manual_extraction_label(
...
-                or token[0] <= token_end_index <= token[1]
-                or token_start_index <= token[0] <= token_end_index
-                or token_start_index <= token[1] <= token_end_index
-            ):
-                return record_item
     new_tokens = record_label_association.create_token_objects(
         project_id, token_start_index, token_end_index + 1
     )
-    record_label_association.create(
+    l = record_label_association.create(
```
FelixKirsch: `l` is never accessed

JWittmeyer: ture :D removed

FelixKirsch (approved):

PR #103, opened 2023-02-02
Implements Gates Update Endpoints

FelixKirsch:
implements endpoints to check if a refinery project is ready to be used with gates and an updater function to make a project gates ready

see gates-gateway for main pr

works with:
https://github.com/code-kern-ai/refinery-ui/pull/115
https://github.com/code-kern-ai/refinery-tokenizer/pull/19
https://github.com/code-kern-ai/refinery-submodule-model/pull/31
https://github.com/code-kern-ai/gates-runtime/pull/3
https://github.com/code-kern-ai/gates-gateway/pull/4
https://github.com/code-kern-ai/gates-ui/pull/3

### DIFF
--- a/controller/project/manager.py
+++ b/controller/project/manager.py
@@ -21,7 +21,7 @@ from submodules.model.business_objects import (
     general,
 )
 from graphql_api.types import HuddleData, ProjectSize, GatesIntegrationData
-from util import daemon
+from util import daemon, notification
 from controller.misc import config_service
 from controller.tokenization.tokenization_service import request_tokenize_project
 from submodules.model.business_objects import util as db_util
@@ -322,6 +322,8 @@ def update_project_for_gates(project_id: str, user_id: str) -> None:
             return
         projects_updating.add(project_id)
 
+    notification.send_organization_update(project_id, "gates_integration:started")
+
     if not __tokenizer_pickle_exists(project_item.tokenizer):
         daemon.run(request_save_tokenizer, project_item.tokenizer)
 
@@ -379,6 +381,8 @@ def __wait_and_create_information_source_pickles(
     with projects_updating_lock:
         if project_id in projects_updating:
             projects_updating.remove(project_id)
+
+    notification.send_organization_update(project_id, "gates_integration:finished")
 
 
 def __wait_for_tokenizer(project_id: str) -> None:
--- a/controller/project/manager.py
+++ b/controller/project/manager.py
@@ -12,6 +12,7 @@ from submodules.model.business_objects import (
     project,
     record_label_association,
     data_slice,
+    embedding,
 )
 from graphql_api.types import HuddleData, ProjectSize
 from util import daemon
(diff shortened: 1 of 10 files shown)


controller/project/manager.py
```diff
@@ -219,12 +229,182 @@ def __get_crowd_label_is_data(project_id: str, is_id: str) -> Dict[str, str]:
...
+
+
+def get_gates_integration_data(project_id: str) -> GatesIntegrationData:
+
+    project_item = project.get(project_id)
+    if not project_item:
+        raise GraphQLError("Project not found")
+
+    if project_id in projects_updating:
+        return GatesIntegrationData(status=enums.GatesIntegrationStatus.UPDATING.value)
```
JWittmeyer, line 254: dont know how perfomrance intensive the missing calculation is but I would have returned them in the updating case as well since with them we might be able to give some king of progression (at a later point in time --> findings)

controller/project/manager.py
```diff
@@ -219,12 +229,182 @@ def __get_crowd_label_is_data(project_id: str, is_id: str) -> Dict[str, str]:
...
+        status=status,
+        missing_tokenizer=missing_tokenizer,
+        missing_embeddings=missing_embeddings,
+        missing_information_sources=missing_information_sources,
+    )
+
+
+def __tokenizer_pickle_exists(config_string: str) -> bool:
+    tokenizer_path = os.path.join(
+        "/inference/tokenizers", f"tokenizer-{config_string}.pkl"
```
JWittmeyer, line 278: is there a specific reason why you are using os.join? i know it should give the correct path seperator for the os but it's fixed to linux in docker anyway (and the path already has some / )

Not wrong just curious if i'd have missed something since i would have gone with `f"/inference/tokenizers/tokenizer-{config_string}.pkl"`

controller/project/manager.py
```diff
@@ -219,12 +229,182 @@ def __get_crowd_label_is_data(project_id: str, is_id: str) -> Dict[str, str]:
...
+def __get_missing_information_source_pickles(project_id: str) -> List[str]:
+    missing = []
+    is_items = information_source.get_all(project_id)
+    for is_item in is_items:
+        if is_item.type != enums.InformationSourceType.ACTIVE_LEARNING.value:
+            # only active learning information sources are pickled
+            continue
+        is_id = str(is_item.id)
+        last_payload = information_source.get_last_payload(project_id, is_id)
+        if last_payload.state == enums.PayloadState.FINISHED.value:
```
JWittmeyer, line 305: if the last one is e.g. the 23rd iteration (and a previous worked) we would still have it or do we delete the al model on error?

Also I'm not sure this is what we intend here. Assuming we only have one active learner that only had error cases so far we would still consider the project gates ready, right? (Findings discussion point)

controller/project/manager.py
```diff
@@ -219,12 +229,182 @@ def __get_crowd_label_is_data(project_id: str, is_id: str) -> Dict[str, str]:
...
+        if project_id in projects_updating:
+            return
+        projects_updating.add(project_id)
+
+    if not __tokenizer_pickle_exists(project_item.tokenizer):
+        daemon.run(request_save_tokenizer, project_item.tokenizer)
+
+    missing_emb_pickles = __get_missing_embedding_pickles(project_id)
+    for embedding_id in missing_emb_pickles:
+        __create_embedding_pickle(project_id, embedding_id, user_id)
```
JWittmeyer: i dont know if this works since token embeddings need the token of the tokenization process before starting, right? I don't see a check for that. 

Also I noticed for the project import embedding creation that simply starting all of them will run into issues for >3 embeddings so maybe we need to queue them as well

controller/project/manager.py
```diff
@@ -219,12 +229,186 @@ def __get_crowd_label_is_data(project_id: str, is_id: str) -> Dict[str, str]:
...
+    user_id: str,
+) -> None:
+    # wait to make sure db entries for embeddings are created
+    time.sleep(5)
+    __wait_for_tokenizer(project_id)
+    __wait_for_embeddings(project_id)
+
+    missing_is_ids = __get_missing_information_source_pickles(project_id)
+    for is_id in missing_is_ids:
+        payload_manager.create_payload(project_id, is_id, user_id)
```
JWittmeyer, line 390: same as with embeddings, might overload if we have multiple big active learners in the project so maybe max 2 at a time?

JWittmeyer (approved):

controller/project/manager.py
FelixKirsch: added to findings

controller/project/manager.py
FelixKirsch: added to findings

controller/project/manager.py
FelixKirsch: As the gates integration does not trigger the tokenization itself, it would be a rare case that the tokenization is running when the gates integration endpoint is triggered. Further, for token embedding, the embedding service itself waits until the tokenization is finished. I don't think that we will run into any issues.

For the embeddings, I added a queue. Only one embedding is run at the same time.

controller/project/manager.py
FelixKirsch: I think it is not as problematic as the embedders. Running all heuristics at the same time is also possible via the UI, so we should also have an eye on that. I added it to the findings.

controller/project/manager.py
JWittmeyer: uploading an old file, seeing "not useable in gates" & clicking on "upgrade" will create the issue for token embeddings

I think the likeleyhood might depend on the users preferences on how they "backup" data -> one way to circumvent it would be to disable the button if an embedding needs tokens & tokenization is still running

controller/project/manager.py
JWittmeyer: yeah there already is an issue for some kind of queue logic for the run all because it often runs into issues :)

PR #104, opened 2023-02-05
create workflow project from REST API instead of GQL

jhoetter:
This PR converts the GraphQL call to create a refinery project from a GraphQL mutation into a REST endpoint, which is then called from the workflow-backend instead of the workflow-frontend

### DIFF
--- a/api/project.py
+++ b/api/project.py
@@ -67,8 +67,11 @@ class ProjectCreationFromWorkflow(HTTPEndpoint):
             organization.id, name, description, user.id
         )
         project_manager.update_project(project_id=project.id, tokenizer=tokenizer)
+        data = adapter.get_records_from_store(store_id)
+        adapter.check(data, project.id, user.id)
+
         project_manager.add_workflow_store_data_to_project(
-            store_id=store_id, user_id=user.id, project_id=project.id, file_name=name
+            user_id=user.id, project_id=project.id, file_name=name, data=data
         )
 
         tokenization_service.request_tokenize_project(str(project.id), str(user.id))
--- a/api/project.py
+++ b/api/project.py
@@ -7,6 +7,29 @@ from controller.project import manager as project_manager
 from controller.attribute import manager as attribute_manager
 from submodules.model import exceptions
 
+import os
+from controller.tokenization import tokenization_service
+from submodules.model import enums, events
+from submodules.s3.controller import bucket_exists, create_bucket
+from util import doc_ock, notification
+from controller.transfer.record_transfer_manager import import_records_and_rlas
+from controller.transfer.manager import check_and_add_running_id
+from controller.auth import manager as auth_manager
+from controller.project import manager as project_manager
+from controller.upload_task import manager as upload_task_manager
+from submodules.model import enums
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+__engine = None
+
+
+def get_workflow_db_engine():
+    global __engine
+    if __engine is None:
+        __engine = create_engine(os.getenv("WORKFLOW_POSTGRES"))
+    return __engine
+
 
 logging.basicConfig(level=logging.DEBUG)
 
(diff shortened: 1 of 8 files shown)


jhoetter:
I also added a bugfix for the `import_json` REST API endpoint

api/project.py
```diff
@@ -7,6 +7,29 @@
...
+from sqlalchemy.orm import sessionmaker
+
+__engine = None
+
+
+def get_workflow_db_engine():
+    global __engine
+    if __engine is None:
+        __engine = create_engine(os.getenv("WORKFLOW_POSTGRES"))
+    return __engine
```
SirDegraf: database connection/engine management seems not optimal placed in an api file, i would expect some sort of layer for this or at least an util file?

api/project.py
```diff
@@ -41,3 +64,62 @@ def get(self, request) -> JSONResponse:
             "knowledge_base_ids": [str(list.id) for list in project.knowledge_bases],
         }
         return JSONResponse(result)
+
+
+class ProjectCreationFromWorkflow(HTTPEndpoint):
```
SirDegraf, line 50: much of what happens here would be outsourced to another layer in our other services. Request managment (reading data out of request and auth of user) can of course be done in an api, but application logic like checking and creating a bucket or session management of a database could be put into a controller file, at least thats what i would have expected

api/project.py
```diff
@@ -41,3 +64,62 @@ def get(self, request) -> JSONResponse:
...
+        project = project_manager.create_project(
+            organization.id, name, description, user.id
+        )
+        project_manager.update_project(project_id=project.id, tokenizer=tokenizer)
+
+        Session = sessionmaker(get_workflow_db_engine())
+        with Session() as session:
+            results = session.execute(
+                f"SELECT record FROM store_entry WHERE store_id = '{store_id}'"
+            ).all()
```
SirDegraf: SQL-Code and its execution should be put in the model submodule

api/project.py
```diff
@@ -41,3 +64,62 @@ def get(self, request) -> JSONResponse:
...
+
+        data = [result for result, in results]
```
SirDegraf: maybe replace var names by more expressive naming

api/project.py
```diff
@@ -41,3 +64,62 @@ def get(self, request) -> JSONResponse:
...
+
+        upload_task = upload_task_manager.create_upload_task(
+            user_id=user.id,
+            project_id=project.id,
+            file_name=name,
+            file_type="json",
```
SirDegraf: use of enum is recommended for constant strings

api/transfer.py
```diff
@@ -132,10 +137,26 @@ async def post(self, request) -> JSONResponse:
         request_body = await request.json()
```
SirDegraf, line 137: again complex application logic in an api

api/transfer.py
```diff
@@ -132,10 +137,26 @@ async def post(self, request) -> JSONResponse:
...
+        for attribute in project.attributes:
+            if attribute.is_primary_key:
+                for idx, record in enumerate(records):
+                    if attribute.name not in record:
+                        if attribute.name == "running_id":
+                            records[idx][attribute.name] = num_project_records + idx + 1
+                        else:
+                            raise exceptions.InvalidInputException(
+                                f"Non-running-id, primary key {attribute.name} missing in record"
+                            )
```
SirDegraf, line 154: Are missing fields allowed when they are not primary keys? I think we forbid it for the "normal" upload but maybe this is a wrong information on my side

app.py
```diff
@@ -40,6 +40,7 @@
     Route("/project/{project_id:str}/import_file", PrepareFileImport),
     Route("/project/{project_id:str}/import_json", JSONImport),
     Route("/project/{project_id:str}/import/task/{task_id:str}", UploadTask),
+    Route("/project", ProjectCreationFromWorkflow),  # create fro workflow
```
SirDegraf:
```suggestion
    Route("/project", ProjectCreationFromWorkflow),  # create from workflow
```

SirDegraf: But since the class is already expressive named, i would remove the comment i think

SirDegraf:
For me it seems like something was changed in the submodule model, too. Maybe I am just missing something, but I would expect a linked submodule PR, but I can't find one. Is there one?

image

api/project.py
jhoetter: created an adapter file in utils for this

api/project.py
jhoetter: this is SQL to access the workflow db, so i will also put it into the adapter

api/project.py
jhoetter: good advice, changed it to:
```python
def get_records_from_store(store_id: str) -> List[Dict[str, Any]]:
    Session = sessionmaker(get_workflow_db_engine())
    with Session() as session:
        record_entity_list = session.execute(
            f"SELECT record FROM store_entry WHERE store_id = '{store_id}'"
        ).all()

    record_dict_list = [result for result, in record_entity_list]
    return record_dict_list
```

api/project.py
jhoetter: added these enums:
```python
class RecordImportFileTypes(Enum):
    JSON = "JSON"
    CSV = "CSV"
    XLSX = "XLSX"

```

api/transfer.py
jhoetter: They should be allowed, just considered as NaN fields. I think this is already happening in refinery, right?

app.py
jhoetter: makes sense, removed it :)

jhoetter:
> For me it seems like something was changed in the submodule model, too. Maybe I am just missing something, but I would expect a linked submodule PR, but I can't find one. Is there one?
> 
> image

yes, that was a mistake on my side, didn't push the submodules. I always forget about them...

api/project.py
jhoetter: where applicable, i transferred it into the controller logic of project (`add_workflow_store_data_to_project`)

api/transfer.py
jhoetter: created a method `unpack_request_body` in the adapter for this

SirDegraf (approved):

PR #111, opened 2023-02-08
Gates Integration Improvements 

FelixKirsch:
use with:

https://github.com/code-kern-ai/refinery-gateway/pull/111
https://github.com/code-kern-ai/refinery-submodule-model/pull/33


JWittmeyer:
maybe we should invalidate the embeddings that are being recalculate so it doesn't look like its already done:
image

(state a second after I clicked on update)


- [x] resolved

JWittmeyer (approved):

PR #119, opened 2023-02-28
Admin features new

SimonDegrafKern:
Gateway https://github.com/code-kern-ai/refinery-gateway/pull/119
Model submodule https://github.com/code-kern-ai/refinery-submodule-model/pull/36
Admin dashboard: https://github.com/code-kern-ai/admin-dashboard/pull/62
Refinery UI: https://github.com/code-kern-ai/refinery-ui/pull/125
Updater https://github.com/code-kern-ai/refinery-updater/pull/28
Config: https://github.com/code-kern-ai/refinery-config/pull/20


controller/admin_message/manager.py
```diff
@@ -0,0 +1,53 @@
+from typing import Any, List
+from datetime import datetime
+from submodules.model.business_objects import admin_message
+from submodules.model.models import AdminMessage
+from util.notification import send_global_update_for_all_organizations
+
+
+def get_all_admin_messages() -> List[AdminMessage]:
+    get_all_active_admin_messages()  # set messages on archived if they should be archived
```
JWittmeyer: imo not optimal regarding naming & function since we dont use the return value. Maybe another (private) function to check/set archived which is used in both (and returns True/False if something was archived)

controller/admin_message/manager.py
```diff
@@ -0,0 +1,53 @@
...
+
+    messages_to_return = []
+    now = datetime.now()
+
+    for message in messages:
+        if now > message.archive_date:
+            admin_message.archive(
+                message.id, None, now, "Archive date reached.", with_commit=True
+            )
+            send_global_update_for_all_organizations(f"admin_message")
```
JWittmeyer: this sends an update for every message that is archived instead of sending one if any is archived.

also f string without interpolation

controller/admin_message/manager.py
```diff
@@ -0,0 +1,53 @@
...
+def get_all_active_admin_messages() -> List[AdminMessage]:
+    messages = admin_message.get_all_active()
+
+    messages_to_return = []
+    now = datetime.now()
+
+    for message in messages:
+        if now > message.archive_date:
+            admin_message.archive(
+                message.id, None, now, "Archive date reached.", with_commit=True
```
JWittmeyer: this commits for every to be archived message, like with the websocket notification I'd go with a "something_changed" flag and only commit once

controller/admin_message/manager.py
```diff
@@ -0,0 +1,53 @@
...
+        else:
+            messages_to_return.append(message)
+
+    return messages_to_return
+
+
+def create_admin_message(
+    text: str, level: str, archive_date: datetime, created_by: str
+) -> AdminMessage:
+    now = datetime.now().astimezone(archive_date.tzinfo)
```
JWittmeyer, line 34: we use different implementations of now over this file. Is this wanted?

controller/admin_message/manager.py
```diff
@@ -0,0 +1,53 @@
...
+
+    if archive_date < now:
+        raise Exception("Archive date not valid")
```
JWittmeyer: Maybe a GraphQLError since this is only called via graphql.

controller/monitor/manager.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any, List
+from submodules.model.business_objects import monitor
+
+
+def monitor_all_tasks(project_id: str = None, only_running: bool = False) -> List[Any]:
```
JWittmeyer: since all tasks is most likely a long list on our app I think I would default this to True to prevent someone from to collect to much if not necessary

controller/monitor/manager.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any, List
+from submodules.model.business_objects import monitor
```
JWittmeyer: monitor is ambiguous imo. Maybe task_monitor?

controller/monitor/manager.py
```diff
@@ -0,0 +1,42 @@
...
+def monitor_all_tasks(project_id: str = None, only_running: bool = False) -> List[Any]:
+    return monitor.get_all_tasks(project_id, only_running)
+
+
+def cancel_all_running_tasks(project_id: str = None) -> None:
+    monitor.cancel_all_running_tasks(project_id)
+
+
+def cancel_upload_task(project_id: str = None, upload_task_id: str = None) -> None:
+    monitor.set_upload_task_to_failed(project_id, upload_task_id, True)
```
JWittmeyer: usually we set with_commit=True (named variable) to prevent misunderstandings & later extension issues

JWittmeyer: same for the other methods in the file

controller/user/manager.py
```diff
@@ -74,3 +79,16 @@ def remove_organization_from_user(user_mail: str) -> None:
         raise Exception("User has no organization")
 
     user.remove_organization(user_id, with_commit=True)
+
+
+def get_active_users(minutes: str, order_by_interaction: bool) -> User:
+    now = datetime.now()
+    last_interaction_range = now - timedelta(minutes=minutes) if minutes else None
```
JWittmeyer: not 100% what the result should be.
if minutes aren't set is last_interaction_range = None or Now?

For readability reasons, I would either use brackets or not the inline syntax

graphql_api/mutation/admin_message.py
```diff
@@ -0,0 +1,43 @@
...
+        archive_date = graphene.DateTime(required=True)
+
+    ok = graphene.Boolean()
+
+    def mutate(self, info, text: str, level: str, archive_date: datetime):
+        auth.check_demo_access(info)
+        auth.check_admin_access(info)
+        user_id = auth.get_user_id_by_info(info)
+        manager.create_admin_message(text, level, archive_date, user_id)
+        send_global_update_for_all_organizations(f"admin_message")
```
JWittmeyer: f string without interpolation

graphql_api/mutation/monitor.py
```diff
@@ -0,0 +1,94 @@
...
+class CancelTokenizationTask(graphene.Mutation):
+    class Arguments:
+        task_id = graphene.ID(required=True)
+
+    ok = graphene.Boolean()
+
+    def mutate(self, info, task_id: str):
+        auth.check_admin_access(info)
+        manager.cancel_record_tokenization_task(tokenization_task_id=task_id)
+        return CancelTokenizationTask(ok=True)
```
JWittmeyer: for easier extension/changes in the future, i'd go with one mutation that get a type, a project_id and a cancel_id. Depending on the type I would then switch between the manager methods.

graphql_api/query/admin_message.py
```diff
@@ -0,0 +1,24 @@
+import graphene
+
+from controller.auth import manager as auth
+from graphql_api.types import AdminMessage
+from controller.admin_message import manager
+
+
+class AdminMessageQuery(graphene.ObjectType):
+    all_admin_messages = graphene.Field(
```
JWittmeyer, line 9: maybe add limit that defaults to 100 (though 100 will take some time until we reach it i'd use it to have a failsafe)

graphql_api/query/monitor.py
```diff
@@ -0,0 +1,12 @@
+import graphene
+from graphql_api.types import Task
+from controller.auth import manager as auth
+from controller.monitor import manager
+
+
+class MonitorQuery(graphene.ObjectType):
+    all_tasks = graphene.Field(graphene.List(Task), only_running=graphene.Boolean())
+
+    def resolve_all_tasks(self, info, only_running: bool = None):
```
JWittmeyer: default True imo

graphql_api/types.py
```diff
@@ -784,3 +783,20 @@ class GatesIntegrationData(graphene.ObjectType):
...
+class AdminMessage(SQLAlchemyObjectType):
+    class Meta:
+        model = models.AdminMessage
+        interfaces = (Node,)
+
+    id = graphene.ID(source="id", required=True)
+
+
+class Task(graphene.ObjectType):
+    id = graphene.String()
```
JWittmeyer: why not graphene.ID() ?

graphql_api/types.py
```diff
@@ -784,3 +783,20 @@ class GatesIntegrationData(graphene.ObjectType):
...
+    organization_name = graphene.String()
+    created_by = graphene.String()
+    project_id = graphene.String()
```
JWittmeyer: graphene.ID()

util/notification.py
```diff
@@ -105,3 +123,13 @@ def get_notification_data(notification_type: str) -> Dict[str, str]:
             f"No data for type {notification_type} found."
         )
     return notification_data
+
+
+def __check_endpoint_set():
```
JWittmeyer: return type

util/notification.py
```diff
@@ -105,3 +123,13 @@ def get_notification_data(notification_type: str) -> Dict[str, str]:
...
+    endpoint = os.getenv("WS_NOTIFY_ENDPOINT")
```
JWittmeyer: we have a const reference for this WEBSOCKET_ENDPOINT. Though I don't know if env variables might change during runtime (but we never checked that on other occasions so i'd go with the const)

controller/admin_message/manager.py
```diff
@@ -0,0 +1,53 @@
...
+
+    return messages_to_return
+
+
+def create_admin_message(
+    text: str, level: str, archive_date: datetime, created_by: str
+) -> AdminMessage:
+    now = datetime.now().astimezone(archive_date.tzinfo)
+
+    if archive_date < now:
```
lumburovskalina, line 36: I think we need to use timestamp function to compare the dates

controller/admin_message/manager.py
SimonDegrafKern: done

controller/admin_message/manager.py
SimonDegrafKern: Done

controller/admin_message/manager.py
SimonDegrafKern: We use the same implementation of now but add for one timezone information because else we could not compare it with the frontend timestamp

controller/admin_message/manager.py
SimonDegrafKern: So in tests this works for me, what benefit would timestamp bring?

controller/admin_message/manager.py
SimonDegrafKern: I don't think this adds any benefit here, maybe something like ValueError since the Date is not fitting in our semantic scope we defined in case of exception

controller/monitor/manager.py
SimonDegrafKern: done

controller/monitor/manager.py
SimonDegrafKern: done

controller/monitor/manager.py
SimonDegrafKern: For the manager methods not I would have thought, in the business object methods this behavior is there

SimonDegrafKern: ah nvm misunderstood the question

controller/user/manager.py
SimonDegrafKern: Done

graphql_api/mutation/admin_message.py
SimonDegrafKern: Done

controller/admin_message/manager.py
JWittmeyer: works for me

controller/admin_message/manager.py
JWittmeyer: ValueError works

controller/admin_message/manager.py
JWittmeyer: my comment doesnt make sense imo - can be resolved/deleted

graphql_api/mutation/monitor.py
SimonDegrafKern: done

graphql_api/query/admin_message.py
SimonDegrafKern: done

graphql_api/query/monitor.py
SimonDegrafKern: Done

graphql_api/types.py
SimonDegrafKern: Done

graphql_api/types.py
SimonDegrafKern: Done

util/notification.py
SimonDegrafKern: Done

util/notification.py
SimonDegrafKern: Dont exactly know what you mean so maybe lets talk about this

controller/admin_message/manager.py
SimonDegrafKern: I understand where you are coming from but I had this specific in mind building this, I renamed the method since the naming really is poor. Lets talk about this and then decide if it should be changed :)

util/notification.py
JWittmeyer: we don't need to get the os env variable with every call of the function. In the notification.py file is a const reference to the endpoint 
image

SimonDegrafKern: Ah ok, yeah this makes sense

JWittmeyer:
I'm not super sure how but I think there is still some issue in the logic since I never touched the users without a role assignment yet they still got a timestamp.
image

Only way I could find to reproduce was.
1. start fresh
2. Create org
3. Assign user to org
4. switch to refinery
5. import first project

That said I tried to find it with throwing an exception on the id but nothing happened at that point so maybe some further investigation is required
image

- [x] resolved

controller/admin_message/manager.py
SimonDegrafKern: Changed it to one function which now has an active_only flag, depending on that only active messages get queried and returned or all, the check and archiving of outdated ones is still there.

graphql_api/types.py
```diff
@@ -784,3 +783,21 @@ class GatesIntegrationData(graphene.ObjectType):
...
+class AdminMessage(SQLAlchemyObjectType):
+    class Meta:
+        model = models.AdminMessage
+        interfaces = (Node,)
+
+    id = graphene.ID(source="id", required=True)
+
+
+class Task(graphene.ObjectType):
+    id = graphene.ID()
```
JWittmeyer, line 797: should we maybe extend this by a start / end date so it's easier to detect "faulty" ones?

JWittmeyer: just checked not all have start/end date so this would be more of an extension issue imo. (so adding start & end date for all types or something but not atm)

util/notification.py
```diff
@@ -20,17 +20,35 @@
 WEBSOCKET_ENDPOINT = os.getenv("WS_NOTIFY_ENDPOINT")
 
 
+def send_global_update_for_all_organizations(message: str):
+    endpoint = os.getenv("WS_NOTIFY_ENDPOINT")
```
JWittmeyer: Endpoint constant

graphql_api/types.py
SimonDegrafKern: Agree, would be really useful but most certainly in a later point of time

util/notification.py
SimonDegrafKern: Done

JWittmeyer (approved):

PR #123, opened 2023-03-13
Attribute calculation for new uploaded records

### LINKED ISSUE #241: Run attribute calculation for newly uploaded records
**Is your feature request related to a problem? Please describe.**
If the user uploads new records and there are usable attributes before the user uploads the records, the import would fail because the attribute calculation is not executed for the newly uploaded records. 

**Describe the solution you'd like**
Start the attribute calculation for the newly uploaded records for all usable attributes. If the attribute calculation fails, the state of that attribute should be set to failed. If there is at least one attribute running, the button 'Upload records' should be disabled. If there are multiple usable attributes, the tokenization should be executed in the end.

lumburovskalina:
- Closes: https://github.com/code-kern-ai/refinery/issues/241

PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/129
- https://github.com/code-kern-ai/refinery-gateway/pull/123
- https://github.com/code-kern-ai/refinery-submodule-model/pull/39
- https://github.com/code-kern-ai/refinery-tokenizer/pull/28


api/transfer.py
```diff
@@ -258,3 +273,161 @@ def file_import_error_handling(
...
+        __calculate_missing_attributes,
+        project_id,
+        user_id,
+    )
+
+
+def __calculate_missing_attributes(project_id: str, user_id: str) -> None:
+    time.sleep(
+        5
+    )  # wait a second to ensure that the process is started in the tokenization service
```
JWittmeyer: comment before the line so it's not in three :)

JWittmeyer (approved):

PR #125, opened 2023-04-27
Adds basic queue

### LINKED ISSUE #183: [BUG] - Queue for execution of lf,al, (zs) & ac
**Describe the bug**
When starting multiple functions in a project (select all & run selected) a lot of containers are created and the functions start erroring out

**To Reproduce**
Steps to reproduce the behavior (depending on your machine):
1. Go to heuristics
2. Click on select all
3. Click on run selected
4. repleat in different project und errors appear

**Expected behavior**
should run through at some point - > best guess a queue with max e.g. 3 containers at the same time

**Additional context**
A execution queue would help here to prevent the docker containers to steal resources from one another.
The limit of the queue is highly dependent on the performance of the system & the size of projects data etc. (usually al is more mem intensive)

Creating a queue will also be a first step to split data further into chunks for "multi threading" in big projects for a single lf

JWittmeyer:
- https://github.com/code-kern-ai/refinery-ui/pull/135
- https://github.com/code-kern-ai/refinery-gateway/pull/125
- https://github.com/code-kern-ai/refinery-submodule-model/pull/40

- closes https://github.com/code-kern-ai/refinery/issues/183
___________________________________
known limits
-> information source & attribute runs are queued without the source code -> if between run and exec a source code change is done this can run into issues
-> gates integration updater has it's own queueing / order logic so nothing changed there atm
-> same for logic after uploaded records

______________________________________
- [ ] add env variables to dev setup & deployments


controller/embedding/manager.py
```diff
@@ -112,3 +113,18 @@ def __embed_one_by_one_helper(
...
         while util.has_encoder_running(project_id):
             time.sleep(5)
+
+
+def get_embedding_name(
+    project_id: str, attribute_id: str, level: str, embedding_handle: str
+) -> str:
+    if level not in ["attribute", "token"]:
+        raise ValueError("level must be either attribute or token")
+    embedding_type = "classification" if level == "attribute" else "extraction"
```
FelixKirschKern: What do you think about introducing enums for the embedding level and the embedding type?

controller/task_queue/manager.py
```diff
@@ -0,0 +1,76 @@
...
+    attribute_calculation as attribute_calculation_handler,
+)
+import copy
+
+from controller.task_queue import task_queue
+
+
+def add_task(
+    project_id: str,
+    type: enums.TaskType,
```
FelixKirschKern: I prefer using `task_type` as variable name because `type` overwrites the builtin function `type()`

controller/task_queue/manager.py
```diff
@@ -0,0 +1,76 @@
...
+    if type == enums.TaskType.INFORMATION_SOURCE.value:
+        return information_source_handler.get_task_functions()
+    if type == enums.TaskType.TOKENIZATION.value:
+        return tokenization_handler.get_task_functions()
+    if type == enums.TaskType.ATTRIBUTE_CALCULATION.value:
+        return attribute_calculation_handler.get_task_functions()
+    raise ValueError(f"Task type {type} not supported yet")
+
+
+def add_task_to_task_queue(task: TaskQueueDBObj):
```
FelixKirschKern: typing `-> None`

controller/task_queue/manager.py
```diff
@@ -0,0 +1,76 @@
...
+    start_func, check_func, check_every = get_task_function_by_type(task.type)
+    task_queue.get_task_queue().add_task(task, start_func, check_func, check_every)
+
+
+def remove_task_from_queue(project_id: str, task_id: str):
```
FelixKirschKern: typing `-> None`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+            return len(self._list)
+
+    def print(self):
+        with self._lock:
+            return print(self._list, flush=True)
+
+
+class CustomTaskQueue:
+    class TaskInfo:
+        def __init__(
```
FelixKirschKern, line 44: typing `-> None`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+            start_function: Callable[[Dict[str, Any]], bool],
+            check_finished_function: Callable[[Dict[str, Any]], bool],
+            check_every: int,
+        ):
+            self.task_dict = manager.parse_task_to_dict(task)
+            self.start_function = start_function
+            self.check_finished_function = check_finished_function
+            self.check_every = check_every
+
+    def __init__(self, max_normal: int, max_priority: int):
```
FelixKirschKern: typing `-> None`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+        )
+        self._checker_thread.start()
+
+    def add_task(
+        self,
+        task: TaskQueueDBObj,
+        start_function: Callable[[Dict[str, Any]], bool],
+        check_finished_function: Callable[[Dict[str, Any]], bool],
+        check_every: int = 1,
+    ):
```
FelixKirschKern: typing `-> None`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+
+        active = self._active_priority if to_prio else self._active_normal
+        # since multiple gateway container can exist the start can "fail" if the task is already active
+        if not task_info.start_function(task_info.task_dict):
+            return False
+        active.append(task_info)
+        task_info.task_dict["is_active"] = True
+        return True
+
+    def __thread_checker(self):
```
FelixKirschKern: typing `-> None`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+                seconds = 0
+                ctx_token = general.remove_and_refresh_session(ctx_token, True)
+            try:
+                self.__check_task_queue(True, seconds)
+                self.__check_task_queue(False, seconds)
+            except:
+                print(traceback.format_exc(), flush=True)
+            time.sleep(1)
+
+    def __check_task_queue(self, priority: bool, seconds: int):
```
FelixKirschKern: typing `-> None`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def append(self, value):
+        with self._lock:
+            self._list.append(value)
+
+    def extend(self, value):
+        with self._lock:
+            self._list.extend(value)
+
+    def pop(self, index=-1):
```
FelixKirschKern: typing `index: int = -1`

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+
+        to_remove = []
+        for idx in range(to_check.length()):
+            task = to_check.get(idx)
+            if seconds % task.check_every == 0:
+                if task.check_finished_function(task.task_dict):
+                    task.task_dict["is_active"] = False
+                    to_remove.append(idx)
+
+        # cleanup
```
FelixKirschKern: The comment is a bit misleading, as not only old tasks are removed but also new ones could be added

controller/task_queue/task_queue.py
```diff
@@ -0,0 +1,193 @@
...
+        # normal queue can solve priority tasks but not the other way around
+        if priority:
+            if self._fifo_queue_priority.length() > 0:
+                return self._fifo_queue_priority.pop(0), True
+        else:
+            if self._fifo_queue_normal.length() > 0:
+                return self._fifo_queue_normal.pop(0), False
+            if self._fifo_queue_priority.length() > 0:
+                return self._fifo_queue_priority.pop(0), False
+        return None, False
```
FelixKirschKern, line 177: For better performance, we could use a `deque` instead of a list for the `fifo` queues.
https://docs.python.org/3/library/collections.html#collections.deque

graphql_api/query/task_queue.py
```diff
@@ -0,0 +1,23 @@
...
+    queued_tasks = graphene.List(
+        TaskQueue,
+        project_id=graphene.ID(required=True),
+        task_type=graphene.String(required=True),
+    )
+
+    def resolve_queued_tasks(
+        self, info, project_id: str, task_type: str
+    ) -> List[TaskQueue]:
+        auth.check_project_access(info, project_id)
```
FelixKirschKern, line 23: I think we also call `auth.check_demo_access` on every query

graphql_api/mutation/embedding.py
```diff
@@ -17,8 +20,22 @@ def mutate(self, info, project_id: str, attribute_id: str, embedding_handle: str
         auth.check_demo_access(info)
         auth.check_project_access(info, project_id)
         user = get_user_by_info(info)
-        manager.create_attribute_level_embedding(
-            project_id, user.id, attribute_id, embedding_handle
+        type = "attribute"
```
FelixKirschKern: we could also use the enum for embedding type here, if we decide to create it

graphql_api/mutation/embedding.py
```diff
@@ -35,8 +52,22 @@ def mutate(self, info, project_id: str, attribute_id: str, embedding_handle: str
         auth.check_demo_access(info)
         auth.check_project_access(info, project_id)
         user = get_user_by_info(info)
-        manager.create_token_level_embedding(
-            project_id, user.id, attribute_id, embedding_handle
+        type = "token"
```
FelixKirschKern: enum for the embedding type, also here

graphql_api/mutation/embedding.py
FelixKirschKern: And I would prefer to call the variable `embedding_type` :)

graphql_api/mutation/embedding.py
FelixKirschKern: And I would prefer to call the variable `embedding_type` :)

controller/embedding/manager.py
JWittmeyer: ill use EmbeddingType (existing enum) instead

controller/task_queue/handler/tokenization.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+    else:
+        tokenization_service.request_tokenize_calculated_attribute(
+            project_id,
+            task["created_by"],
+            task["task_info"]["attribute_id"],
+            task["task_info"]["include_rats"],
+        )
+
+    return True
```
FelixKirschKern: If the tokenization of a calculated attribute is the task, we should check if the attribute is still existing, otherwise the request to the tokenization service runs into errors.

controller/task_queue/task_queue.py
JWittmeyer: changed to `# cleanup & requeue existing tasks`

graphql_api/mutation/embedding.py
JWittmeyer: done

graphql_api/mutation/embedding.py
JWittmeyer: done

controller/task_queue/task_queue.py
JWittmeyer: as discussed list will be used

controller/task_queue/handler/tokenization.py
```diff
@@ -19,10 +21,20 @@ def __start_task(task: Dict[str, Any]) -> bool:
...
     project_id = task["project_id"]
 
+    if task["task_info"]["scope"] == RecordTokenizationScope.ATTRIBUTE.value:
+        attribute_item = attribute_db_bo.get(
+            project_id, task["task_info"]["attribute_id"]
+        )
+        if attribute_item is None or (
+            attribute_item.state
+            not in [AttributeState.INITIAL.value, AttributeState.FAILED.value]
+        ):
```
FelixKirschKern: This condition leads to an error because the calculated attribute is set to state usable directly after the tokenization is requested. So at the time when the tokenization is executed, the attribute is already in state usable.
I suggest to only check if the `attribute_item` is `None` and omitting the check for the states.

controller/task_queue/manager.py
```diff
@@ -0,0 +1,76 @@
...
+    user_id: str,
+    task_info: Dict[str, str],
+    priority: bool = False,
+) -> str:
+    task_item = task_queue_db_bo.add(
+        project_id, task_type, user_id, task_info, priority, with_commit=True
+    )
+
+    add_task_to_task_queue(task_item)
+    return str(task_item.id)
```
FelixKirschKern: It could be that the `task_item` is deleted in the function call of `add_task_to_task_queue`. If nothing is in the queue the task is started directly and deleted if some preconditions are not fulfilled. In this case, we run into an error when accessing `task_item.id`.

controller/task_queue/handler/tokenization.py
JWittmeyer: changed to not USEABLE

controller/task_queue/manager.py
JWittmeyer: stored beforehand to prevent error

FelixKirschKern (approved):

PR #129, opened 2023-05-16
Adds progress update to execution env

SimonDegrafKern:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/137
- https://github.com/code-kern-ai/refinery-gateway/pull/129
- https://github.com/code-kern-ai/refinery-submodule-model/pull/42
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/19


controller/attribute/util.py
```diff
@@ -81,16 +83,18 @@ def run_attribute_calculation_exec_env(
...
 
-    attribute.update(
-        project_id=project_id, attribute_id=attribute_id, logs=logs, with_commit=True
-    )
+        splitted_line = line.split(":")
+        if len(splitted_line) > 1:
+            if splitted_line[-2] == "progress":
+                __update_progress(project_id=project_id, attribute_id=attribute_id, logs=logs, progress=round(float(splitted_line[-1]),2))
+        
+    __update_progress(project_id=project_id, attribute_id=attribute_id, logs=logs, progress=1.0)
```
JWittmeyer: does this work? for lf exec env I needed to run a thread to access the logs (read_container_logs_thread) since collecting the logs does it only after the container is completed which would result in the progress jumping from 0 to 100?

controller/attribute/util.py
```diff
@@ -106,3 +110,14 @@ def run_attribute_calculation_exec_env(
     s3.delete_object(org_id, project_id + "/" + prefixed_payload)
 
     return calculated_attributes
+
+
+def __update_progress(project_id: str, attribute_id: str, logs: List[str], progress: float) -> None:
+    progress = progress
```
JWittmeyer: what is this for?

JWittmeyer:
Progress isn't filtered from logs:
image

- [x] resolved

Weird, again, just not enough care, happens way to often in the last time, so will change this ofc

controller/attribute/util.py
JWittmeyer: doesn't work. 
Why was decided to go with a custom solution when the task was to do this like lf exec env progress?

SimonDegrafKern: Tbh I think it was just not implemented with enough care from my side because last sprint was full of different todos so my focus maybe went off, so there is no real rational reason but more of just bad work done i guess

controller/attribute/util.py
SimonDegrafKern: nothing i guess, will remove it

controller/attribute/util.py
SimonDegrafKern: will refactor it with a threaded solution

controller/attribute/util.py
```diff
@@ -73,24 +83,29 @@ def run_attribute_calculation_exec_env(
...
             stream=True, stdout=True, stderr=True, timestamps=True
         )
+        if "progress" not in line.decode("utf-8")
     ]
-
-    attribute.update(
-        project_id=project_id, attribute_id=attribute_id, logs=logs, with_commit=True
-    )
+    del __containers_running[container_name]
+    __update_progress(project_id, attribute_item, 1.0)
```
JWittmeyer: the attribute calculation isn't finished with the exec environment. Thats why in the lf logic we set it to 0.05 before starting the container, 0.9 after the container is finished and every progress from the container (0-1) multiplied with 0.8. Why was decided to skip this part here? For me now the progress bar looks finished (with 100%) but still takes quite some time to get the calculated valued to the database.

controller/attribute/util.py
```diff
@@ -106,3 +121,47 @@ def run_attribute_calculation_exec_env(
...
+        if c > 100:
+            ctx_token = general.remove_and_refresh_session(ctx_token, True)
+            attribute_item = attribute.get(project_id, attribute_id)
+        if not container_name in __containers_running:
+            break
+        try:
+            last_progress = 0.0
+            for log in docker_container.logs(
+                stream=True,
+                tail=10
```
JWittmeyer: during my testing for lf progress, I noticed that tail doesn't really work well with user logs (so if someone prints something in the log on their own). So e.g., if someone prints more than 10 between the progress logs, we won't be able to update. 
Further the logs seem to be doubled? 
image


Why was decided to change this from the original logic with the timestamp?

controller/attribute/util.py
SimonDegrafKern: I did not skip the part, I just built a threaded solution which sends the progress of the execution env from 0.0 to 1.0. But ofc I can change it to the behavior you described

controller/attribute/util.py
SimonDegrafKern: Ofc this can lead to problems when someone prints, I could just extend the size of the tail, but there could still be edge cases in which the progress logs will not be collected, but this is imo no real problem since not every progress step is necessary for a fully functional progress overview. 

the timestamp logic is quite complicated to first understand and then refactor imo, so I thought about another solution and came up with this

JWittmeyer (approved):

PR #131, opened 2023-05-16
Update installed versions to remote

lumburovskalina:
PRs: 
- https://github.com/code-kern-ai/refinery-gateway/pull/131
- https://github.com/code-kern-ai/refinery-updater/pull/36
- https://github.com/code-kern-ai/refinery-submodule-model/pull/43


controller/misc/manager.py
```diff
@@ -47,6 +47,9 @@ def get_version_overview() -> List[ServiceVersionResult]:
 def has_updates() -> List[ServiceVersionResult]:
     return __updater_has_updates()
 
+def update_versions_to_newest() -> None:
```
JWittmeyer, line 51: a comment to ensure that everyone know that is only sets the values in the database not the actual update logic (like in the updater service but here as well :)

graphql_api/mutation/misc.py
```diff
@@ -51,8 +51,17 @@ def mutate(self, info, event_name: str, event_data: Dict[str, Any]):
             setattr(event_instance, key, event_data[key])
         doc_ock.post_event(str(user.id), event_instance)
         return UpdateConfig(ok=True)
+    
+class UpdateVersionsToNewest(graphene.Mutation):
+    ok = graphene.Boolean()
+
+    def mutate(self, info):
+        auth.check_demo_access(info)
+        manager.update_versions_to_newest()
```
JWittmeyer: i think we should also check admin access so not everyone could in theory do this 
`auth.check_admin_access(info)`

JWittmeyer (approved):

PR #136, opened 2023-06-02
Embedding providers

SimonDegrafKern:
gateway: https://github.com/code-kern-ai/refinery-gateway/pull/136
embedder: https://github.com/code-kern-ai/refinery-embedder/pull/42
gates: https://github.com/code-kern-ai/gates-runtime/pull/24
submodule model: https://github.com/code-kern-ai/refinery-submodule-model/pull/44
ui: https://github.com/code-kern-ai/refinery-ui/pull/143
admin-dashboard: https://github.com/code-kern-ai/admin-dashboard/pull/8
updater: https://github.com/code-kern-ai/refinery-updater/pull/43


JWittmeyer:
wouldn't the base branch be new-embedding-providers instead of dev?

- [ ] resolved

controller/embedding/manager.py
```diff
@@ -1,18 +1,18 @@
...
-from submodules.model.business_objects import attribute
+from controller.project import manager as project_manager
+from submodules.model.business_objects import attribute, embedding, agreement, general
 
 
-def get_recommended_encoders() -> List[Any]:
+def get_recommended_encoders(is_managed: bool) -> List[Any]:
     recommendations = connector.request_listing_recommended_encoders()
-    if misc.check_is_managed():
+    if is_managed:
```
JWittmeyer, line 17: Since the model provider doesn't exist for os we would run into an error if someone would call the function with true even though it's actually not managed. so imo the parameter would be a "false/faulty" option for a dev. Since the user of the function can't actually decide the call is inside the function.

now the misc call is done in the line before the function.

the functionality for the existing case wasn't changed so why was it changed to a parameter? Both approaches should work but I'm still curious :D

controller/embedding/manager.py
```diff
@@ -96,36 +82,25 @@ def __embed_one_by_one_helper(
...
-                config_string=splitted[2],
-            )
-        elif embedding_item.get("type") == enums.EmbeddingType.ON_TOKEN.value:
-            connector.request_creating_token_level_embedding(
-                project_id,
-                attribute_id=attribute_names[splitted[0]],
-                user_id=user_id,
-                config_string=splitted[2],
-            )
+        platform = project_manager.__get_platform_name(embedding_item.get("name"))
```
JWittmeyer: even though possible, i don't think we should call "private" methods. I'd change it to a public method. (also I couldn't find the function in the gateway so maybe there is something wrong here anyway)

controller/embedding/manager.py
```diff
@@ -134,4 +109,13 @@ def get_embedding_name(
...
     attribute_name = attribute_item.name
 
-    return f"{attribute_name}-{embedding_type}-{embedding_handle}"
+    name = f"{attribute_name}-{embedding_type}-{platform}"
+
+    if model:
+        name += f"-{model}"
+
+    if apiToken:
+        name += f"-{apiToken[:8]}..."
```
JWittmeyer: haven't tested how big of an overlapp there is between key (so e.g. if the first 10 char always describe the organization and the rest something else this might have some issues).

I couldn't find a pattern or anything  listed somewhere but I found how openai displays them

image

so maybe `apiToken[:3]+"..."+apiToken[-4:]`

controller/embedding/util.py
```diff
@@ -1,5 +1,34 @@
...
 
 def has_encoder_running(project_id: str) -> bool:
     return embedding.get_first_running_encoder(project_id) is not None
+
+
+def recreate_embedding(
+    project_id: str, embedding_id: str
+) -> None:
+    old_embedding_item = embedding.get(project_id, embedding_id)
+    agreement_item = agreement.get(project_id, old_embedding_item.agreement_id)
```
JWittmeyer: what about ones that don't have an agreement item? (e.g. hugginface wont have one)
Also on import this agreement item will not exist if I remember correctly we want to throw an error for them and set the old one to error.

controller/embedding/util.py
```diff
@@ -1,5 +1,34 @@
...
+    new_embedding_item = embedding.create(
+        project_id,
+        old_embedding_item.attribute_id,
+        old_embedding_item.name,
+        enums.EmbeddingState.INITIALIZING.value,
```
JWittmeyer: is this correct? maybe i misunderstand the method. I though recreate_embedding can be called from bacially everywhere and it handles the whole recreation thing. If it's just for the upload of new records/gates upgrade (so not task queue related) then we should note that somewhere.

controller/embedding/util.py
```diff
@@ -1,5 +1,34 @@
...
+        type=old_embedding_item.type,
+        model=old_embedding_item.model,
+        platform=old_embedding_item.platform,
+        api_token=old_embedding_item.api_token,
+        with_commit=False
+    )
+    agreement_item.xfkey = new_embedding_item.id
+    general.commit()
+    connector.request_deleting_embedding(project_id, old_embedding_item.id)
```
JWittmeyer: usually we delete first (after collection of necessary information) and create the db entry then so at no point in time the "same" embedding is listed twice in db.

controller/task_queue/handler/embedding.py
```diff
@@ -34,15 +37,30 @@ def __start_task(task: Dict[str, Any]) -> bool:
...
-        )
-    else:
-        embedding_manager.create_token_level_embedding(
-            project_id, user_id, attribute_id, embedding_handle
-        )
+    embedding_type = task["task_info"]["embedding_type"]
+    embedding_name = task["task_info"]["embedding_name"]
+    platform = task["task_info"]["platform"]
+    model = task["task_info"]["model"]
+    api_token = task["task_info"]["api_token"]
```
JWittmeyer, line 44: since we don'T always have an api_token/model I was expecting some condition logic or a dictionary get

controller/task_queue/handler/embedding.py
```diff
@@ -34,15 +37,30 @@ def __start_task(task: Dict[str, Any]) -> bool:
...
+
+    terms_text = task["task_info"]["terms_text"]
+    terms_accepted = task["task_info"]["terms_accepted"]
```
JWittmeyer, line 47: see above

controller/task_queue/handler/embedding.py
```diff
@@ -34,15 +37,30 @@ def __start_task(task: Dict[str, Any]) -> bool:
...
+        attribute_id,
+        embedding_name,
+        enums.EmbeddingState.INITIALIZING.value,
+        type=embedding_type,
+        model=model,
+        platform=platform,
+        api_token=api_token,
+
+    )
+    agreement_db_bo.create(project_id, user_id, terms_text, terms_accepted, xfkey=embedding_item.id, xftype=enums.AgreementType.EMBEDDING.value, with_commit=True)
```
JWittmeyer: not all embedings get an agreement entry, only cohere & openai

controller/transfer/project_transfer_manager.py
```diff
@@ -650,6 +650,12 @@ def import_file(
                 finished_at=embedding_item.get(
                     "finished_at",
                 ),
+                platform=embedding_item.get(
+                    "platform",
+                ),
+                model=embedding_item.get(
+                    "model",
+                ),
```
JWittmeyer: maybe a comment that because of security reasons API key aren't exported so we don't "forget" about it

controller/task_queue/handler/embedding.py
JWittmeyer: sorry misunderstanding the get is done in the mutation -> can be resovled

controller/task_queue/handler/embedding.py
JWittmeyer: or maybe only if a text and accepted != None?

graphql_api/query/embedding.py
```diff
@@ -49,3 +52,27 @@ def resolve_project_tokenization(
                 progress=-1,
             )
         return tokenization.get_record_tokenization_task(project_id)
+
+    def resolve_embedding_platforms(self, info) -> List[EmbeddingPlatform]:
+        return [
```
JWittmeyer, line 58: wouldn't this be something for the embedding repo that is passed through like the suggestions?

controller/embedding/manager.py
SimonDegrafKern: wrote comment

controller/embedding/manager.py
SimonDegrafKern: outdated

controller/embedding/util.py
SimonDegrafKern: Recreate is already solved

Creation of agreements for hugging and python can be excluded/filtered

controller/embedding/util.py
SimonDegrafKern: was refactored and is now to be used in a more generic sense

controller/embedding/util.py
SimonDegrafKern: Solved

controller/transfer/project_transfer_manager.py
SimonDegrafKern: done

graphql_api/query/embedding.py
SimonDegrafKern: done

controller/embedding/manager.py
SimonDegrafKern: solved

controller/task_queue/handler/embedding.py
SimonDegrafKern: solved

controller/embedding/util.py
SimonDegrafKern: solved

api/transfer.py
```diff
@@ -234,7 +236,12 @@ def init_file_import(task: UploadTask, project_id: str, is_global_update: bool)
             import_preperator.prepare_label_studio_import(project_id, task)
         else:
             transfer_manager.import_records_from_file(project_id, task)
-        calculate_missing_attributes(project_id, task.user_id)
+        daemon.run(
+            __recalculate_missing_attributes_and_embeddings,
+            project_id,
+            task.user_id,
```
JWittmeyer: not sure if necessary but task.user_id should be a uuid so maybe a str cast is necessary since the recalculate expects a string.

JWittmeyer:
Can't start a container of the branch, seems the submodule isn't up to date in the build/uses the dev branch for the drone build.

Log:
image

Done submodule collection:
image

commit non dev:
image
commit dev:
image

so far we haven't noticed it because the container aren't used often :D 


(not a mistake or something we want to solve now but an observation because I wanted to not build everything on my computer :D )

controller/embedding/manager.py
```diff
@@ -134,4 +104,110 @@ def get_embedding_name(
...
     attribute_name = attribute_item.name
 
-    return f"{attribute_name}-{embedding_type}-{embedding_handle}"
+    name = f"{attribute_name}-{embedding_type}-{platform}"
+
+    if model:
+        name += f"-{model}"
+
+    if apiToken:
+        name += f"-{apiToken[:8]}...{apiToken[-4:]}"
```
JWittmeyer: this is not the code snipped we discussed `apiToken[:3]+"..."+apiToken[-4:]` (first 3 chars not 8)

controller/embedding/manager.py
```diff
@@ -134,4 +104,110 @@ def get_embedding_name(
...
+
+def __recreate_embedding(
+    project_id: str, embedding_id: str
+) -> Embedding:
+    old_embedding_item = embedding.get(project_id, embedding_id)
+    old_id = old_embedding_item.id
+    new_embedding_item = embedding.create(
+        project_id,
+        old_embedding_item.attribute_id,
+        old_embedding_item.name,
```
JWittmeyer, line 187: this looks to be a copy of controller/embedding/util.py -> recreate_embedding. Didn't check every line so maybe I'm missing something but I think if possible we should reuse the recreate_embedding method

controller/transfer/project_transfer_manager.py
```diff
@@ -611,14 +612,33 @@ def import_file(
...
+        splitted_name = embedding_name.split("-")
+        attribute_name = splitted_name[0]
+        embedding_type = splitted_name[1]
+        model = "-".join(splitted_name[2:])
+        if "bag-of-words" == model or "bag-of-characters" == model or "tf-idf" == model: 
+            platform= enums.EmbeddingPlatform.PYTHON.value
+        else:
+            platform = enums.EmbeddingPlatform.HUGGINGFACE.value
+        name = f"{attribute_name}-{embedding_type}-{platform}-{model}"
+        return platform, model, name
```
JWittmeyer: why is this a function inside a function?

controller/transfer/project_transfer_manager.py
```diff
@@ -850,19 +880,23 @@ def import_file(
...
+
+def __post_processing_import_threaded(project_id: str, task_id: str,  embedding_ids: List[str], data: Dict[str, Any]) -> None:
+    time.sleep(5)
+    while True:
+        if task_queue.get_by_tokenization(project_id):
+            logger.info(f"Waiting for tokenization of project {project_id}")
+            time.sleep(5)
+        else:
+            logger.info(f"Tokenization finished, continue with embedding handling of project {project_id}")
+            break
```
JWittmeyer, line 919: which this logic we now wait for tokenization for all embeddings. With the previous version (though faulty for token specific ones) we didn't wait.

Since only for token specific ones need the tokenization, I think we should do this in two parts (token embedding + waiting & attribute embedding + no waiting).

controller/embedding/manager.py
```diff
@@ -134,4 +104,110 @@ def get_embedding_name(
...
+    for embedding_id in embedding_ids:
+        embedding.update_embedding_state_waiting(project_id, embedding_id)
+        notification.send_organization_update(
+                project_id,
+                f"embedding:{embedding_id}:progress:{0.0}",
+            )
+        notification.send_organization_update(
+                project_id,
+                f"embedding:{embedding_id}:state:{enums.EmbeddingState.WAITING.value}",
+            )
```
JWittmeyer: image

In the frontend on every WAITING state the embeddings query is refetched. This in turn results in two issues. 
1. At the instant WAITING is sent the changes aren't committed yet. (not very critical since some delay in frontend but still possibly problematic)
2. We refetch n times very quickling

i think after the commit one notification with waiting would be enough (with all for the embedding id part)

controller/embedding/manager.py
SimonDegrafKern: Because we do not only use openAI but cohere too I thought we may should have some more tokens in it but yeah can make a 3 out of the 8 if you would prefer it

controller/transfer/project_transfer_manager.py
SimonDegrafKern: It is only used at this place

controller/embedding/manager.py
SimonDegrafKern: This is indeed a new version of the one in util which is not used anymore but was not removed, i removed the one from util

controller/transfer/project_transfer_manager.py
SimonDegrafKern: Since this is not part of the task queue logic yet I would still keep it as this is so we do not run to much in parallel. As soon as we include our queue task manager here I would agree that we can resolve it by creating a dependency between different task types (ON_TOKEN Embeddings have an dependency on Tokenization; ON_ATTRIBUTE do not have it).

controller/embedding/manager.py
SimonDegrafKern: totally agree, will change the call logic

controller/transfer/project_transfer_manager.py
```diff
@@ -287,11 +288,162 @@ def import_file(
...
+                        "data",
+                    ),
+                )
+
+    def __replace_embedding_name(soure_code: str, embedding_name_mapping: Dict[str, str]) -> str:
+        for embedding_name in embedding_name_mapping.keys():
+            if embedding_name in soure_code:
+                code =  soure_code.replace(
+                    embedding_name,
+                    embedding_name_mapping[embedding_name])
```
JWittmeyer: i dont think this is enough. since an embedding name could be partially included so e.g. 

embedding_name = "ABC"

embedding_name = "ABCD"

the direct replace would replace both which can result in errors. I first thought regex word bounds might do the trick but a dash (-) is also a word boundarie so I'd run it twice with
"<embedding_name>" 
'<embedding_name>' 

that way we should be relatively sure that we don't replace something we don't want to.

SimonDegrafKern: How exactly could a embedding name be partially included? 

the old ones are build like this


attribute_name-embedding_type-model_name

e.g.

headline-classification-roberta-base

Maybe I am not thinking hard enough, but what example would there be where one embedding name is included in another one?

SimonDegrafKern: Now that I think about it maybe the attribute part could be problematic e.g.

headline-classification-roberta-base

custom_headline-classification-roberta-base

--> will change it

api/transfer.py
SimonDegrafKern: Added str()

controller/transfer/project_transfer_manager.py
SimonDegrafKern: changed it

controller/embedding/manager.py
SimonDegrafKern: Changed

controller/transfer/project_transfer_manager.py
SimonDegrafKern: Not changed

controller/transfer/project_transfer_manager.py
```diff
@@ -287,11 +288,169 @@ def import_file(
...
+                code =  soure_code.replace(
+                    double_quoted_name,
+                    f'"{embedding_name_mapping[embedding_name]}"')
+                return code
+            elif single_quoted_name in soure_code:
+                code =  soure_code.replace(
+                    single_quoted_name,
+                    f"'{embedding_name_mapping[embedding_name]}'")
+                return code
+        return soure_code
```
JWittmeyer: this could result in an issue if both exist or more than one embedding name is present (not sure how likely but since customer code at least possible).
So maybe without return and elif = if
```suggestion
    def __replace_embedding_name(soure_code: str, embedding_name_mapping: Dict[str, str]) -> str:
        code = source_code
        for embedding_name in embedding_name_mapping.keys():
            double_quoted_name = f'"{embedding_name}"'
            single_quoted_name = f"'{embedding_name}'"
            if double_quoted_name in code:
                code =  code.replace(
                    double_quoted_name,
                    f'"{embedding_name_mapping[embedding_name]}"')
            if single_quoted_name in soure_code:
                code =  code.replace(
                    single_quoted_name,
                    f"'{embedding_name_mapping[embedding_name]}'")
        return code 
```

JWittmeyer:
Import now doesn't work anymore.
image

Code wasn't tested. after suggestion was used. 
Error seems to be a typo in the signature

- [x] resolved

JWittmeyer (approved):

PR #137, opened 2023-06-19
Adds password support for upload task

SimonDegrafKern:
PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/137
- https://github.com/code-kern-ai/refinery-submodule-model/pull/45
- https://github.com/code-kern-ai/refinery-ui/pull/145
- https://github.com/code-kern-ai/dev-setup/pull/30

Additional note:
Secret key added in deployment repos
https://github.com/code-kern-ai/deployment-managed-app/pull/12
https://github.com/code-kern-ai/deployment-demo-app/pull/6


start
```diff
@@ -34,7 +34,7 @@ then
 fi
 
 echo -ne 'building container...'
-docker build -t graphql-dev -f dev.Dockerfile . > /dev/null 2>&1
+docker build -t graphql-dev -f dev.Dockerfile .
```
SimonDegrafKern:
```suggestion
docker build -t graphql-dev -f dev.Dockerfile .  > /dev/null 2>&1
```

start
```diff
@@ -34,7 +34,7 @@ then
 fi
 
 echo -ne 'building container...'
-docker build -t graphql-dev -f dev.Dockerfile . > /dev/null 2>&1
+docker build -t graphql-dev -f dev.Dockerfile .  > /dev/null 2>&1
```
SimonDegrafKern:
```suggestion
docker build -t graphql-dev -f dev.Dockerfile . > /dev/null 2>&1
```

SimonDegrafKern:
@JWittmeyer There can be some weird behavior of the progress bar or redirect on snapshot upload which we traced back to probably be resulting as a side effect of the throttled update of progresses.

@param_throttle(seconds=2)
def send_progress_update_throttle(project_id: str, task_id: str, value: float) -> None:
    send_progress_update(project_id, task_id, value)
    
    
When reviewing we should maybe have a discussion for this

____________________________
had the same issue. 
image
At some point this was working by using a none throttled "final" message so the complete is always send and not throttled away. 

Not sure when it was removed but I'd solve it like this again 😁 
image

____________________________
Check a bit further. The "I'm done with the upload message" still exists but it is send after the embedding "upgrade/recalculation". Since this isn't part of the file upload I would put it before the post processing tasks. 

Simon: Done

Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM kernai/refinery-parent-images:v1.11.0-common
 
 WORKDIR /app
 
+RUN apt-get update && apt-get install -y libc6-dev zlib1g gcc --no-install-recommends
```
JWittmeyer, line 6: Is there a specific reason this is done via the Dockerfile and not the requirements.txt? (They are linux packages vs python packages, I ask why we don't use python packages)

Also libc6-dev sounds like a dev version of a package. Is this the correct one?

controller/transfer/manager.py
```diff
@@ -57,9 +57,11 @@ def get_upload_credentials_and_id(
     file_type: str,
     file_import_options: str,
     upload_type: str,
-):
+    key: str,
```
JWittmeyer: wouldn't this be optional or am i missing something?

start
```diff
@@ -75,6 +75,7 @@ docker run -d --rm \
 -e TASK_QUEUE_SLOTS=1 \
 -e PRIORITY_TASK_QUEUE_SLOTS=1 \
 -e INFERENCE_DIR=$INFERENCE_DIR \
+-e SECRET_KEY=default \
```
JWittmeyer, line 78: so we don't forget about it let's add a note to the `https://github.com/code-kern-ai/deployment-managed-app/pull/12` that we need to add the env variable

util/clean_up.py
```diff
@@ -0,0 +1,19 @@
+from submodules.model.business_objects import upload_task
+import os, shutil
+
+def clean_up_database() -> None:
+    upload_task.remove_all_keys(with_commit=True)
+
+def clean_up_disk() -> None:
```
JWittmeyer, line 9: Do i understand it right das clean_up_disk only checks the tmp folder? Maybe we should add this in the method name since I first thought I need to worry about all files that could be erased :D

util/file.py
```diff
@@ -0,0 +1,43 @@
...
+    try:
+        data = json.loads(zip_file.read(file_name, key).decode())
+        return data
+    except RuntimeError as error:
+        if "password" in str(error):
+            raise BadPasswordError
+        
+def zip_to_json_file(zip_file_path: str, key: Optional[str] = None) -> str:
+    json_data = zip_to_json(zip_file_path, key)
+    file_name = f"{zip_file_path}{random.randint(1,999)}.json"
```
JWittmeyer: why are we using a random integer in the file name? Usually i'd expect a while loop that checks existing and return the first free filename

util/security.py
```diff
@@ -0,0 +1,36 @@
...
+
+SECRET_KEY = os.getenv("SECRET_KEY")
+
+__crypto_suite = None
+
+
+def encrypt(value: str) -> ByteString:
+    if not value:
+        return None
+    global __crypto_suite
```
JWittmeyer: when we have a getter we don't need the global reference

util/security.py
```diff
@@ -0,0 +1,36 @@
...
+    value = __get_crypto_suite().encrypt(value, SECRET_KEY)
+    return value
+
+def decrypt(value: ByteString) -> str:
+    if not value:
+        return None
+    global __crypto_suite
```
JWittmeyer: see above

util/security.py
```diff
@@ -0,0 +1,36 @@
+
+import os
+import rncryptor
+
+from typing import ByteString
+
+SECRET_KEY = os.getenv("SECRET_KEY")
```
JWittmeyer, line 7: should we maybe add a `, "default"` so if someone forgets to add it in the env file / variables we would still have some encryption?

JWittmeyer:
Not 100% since there might have been a different issue but I think the snapshot upload doesn't clear the password after it is done with it?
image

- [ ] resolved

I cant reproduce it on my system, did it happen for you once or does this bug happen multiple times?

Dockerfile
SimonDegrafKern: So these are used because of pyminizip, which itself needs for some reason gcc and these "lib"-somewhat libraries. After testing some different solutions which did not work this was the one which worked. While building some dependencies we did not have a c compiler in our building process and it just died all the time. Can be that this is not the 100% correct scenario since I am really not experienced in linux package management but this is the rough direction of why we did it like this.

controller/transfer/manager.py
SimonDegrafKern: Yeah should be optional

start
SimonDegrafKern: yes right I will do this

util/clean_up.py
SimonDegrafKern: Exactly, it checks and cleans up only the tmp folder. I wrote it with this name so if we want to have further clean up steps they could just be integrated in this method, but of course I can add something like "tmp" in the name of the method

util/file.py
SimonDegrafKern: I mean, both work fine, but of course I can rewrite it

util/security.py
SimonDegrafKern: yes you are right!

Dockerfile
JWittmeyer: works for me, could you then add a comment that they are needed for the encryption stuff then this can be resolved

util/file.py
JWittmeyer: since we don't check the file name this would (of course only in rare instances) create an issue with the same file name being used for two different tasks. to be honest with our current size I don't expect this to ever happen. However, this is not a solution that works 100% of the time and might result in very hard-to-find (and debug) issues.

I would be fine-er with it if it's wrapped in a loop to check the file name.

To solve this for future use as well i'd create a small function like `get_free_file_name(path:str,file_name:str, increment:int = 1):`

Dockerfile
SimonDegrafKern: yes will do it!

util/file.py
SimonDegrafKern: I agree, your way is more fail safe, will quickly rewrite it

start
JWittmeyer: now the secret key for managed is also default. I think it should be different for our applications

util/file.py
```diff
@@ -0,0 +1,52 @@
...
+    with open(json_file_path, "w") as f:
+        f.write(text)
+    return json_file_path
+
+
+def __get_free_file_path(file_path: str, increment: int = 1):
+    if os.path.exists(file_path):
+        file_path = f"{file_path}_{increment}"
+        return __get_free_file_path(file_path, increment + 1)
+    return file_path
```
JWittmeyer: not sure this is what was intended but wouldn't the file name now be {some_path}_1_2_3_4 and so on if they already exist?

I would go with a solution that isn't recursive 

```suggestion
def __get_free_file_path(file_path: str, increment: int = 1):
    check_path = file_path
    counter = 0
    while os.path.exists(check_path):
        counter = counter + increment
        check_path =  f"{file_path}_{counter}"
    return check_path 
```
The suggestion might have some whitespace issues :D

JWittmeyer:
Since this is an "expected error" i wouldn't put it into the log. 
image
if we want we could still print something but the full stacktrace looks like something went wrong even though it's expected with a wrong password
- [x] resolved

util/file.py
SimonDegrafKern: changed

start
SimonDegrafKern: Set to variable and set on managed server a secret key in env file

JWittmeyer (approved):

PR #140, opened 2023-07-13
Custom project name for project snapshot and sample projects

### LINKED ISSUE #261: Project snapshot name can be set when uploading a new snapshot
**Is your feature request related to a problem? Please describe.**
If we want to create a new project with uploading new records, we have a validation that we cannot use the same project name twice, but we export the project and import it again then we have projects with the same name. I noticed that in gates it’s hard to find your project since imported projects and sample projects can have the same name.

**Describe the solution you'd like**
We can have an optional input field in the upload component for snapshots (and maybe lookup lists) where the user can enter the desired name (same validation as we have when creating a new project with records).

**Describe alternatives you've considered**
 We could also think for a differentiation of the sample projects names (index of the sample project, or entering a name before submitting the project..).

lumburovskalina:
Closes: https://github.com/code-kern-ai/refinery/issues/261
PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/140
- https://github.com/code-kern-ai/refinery-ui/pull/149
- https://github.com/code-kern-ai/refinery-submodule-model/pull/47


controller/transfer/project_transfer_manager.py
```diff
@@ -82,19 +82,19 @@ def ensure_project_exists(project_id: str) -> None:
...
+    elif project_type == enums.SampleProjectType.AG_NEWS.value:
         file_name = "sample_projects/ag_news.zip"
-    elif project_name == "Conversational AI - initial":
+    elif project_type == enums.SampleProjectType.CONVERSATIONAL_AI_INITIAL.value:
         file_name = "sample_projects/conversational_ai_initial.zip"
-    elif project_name == "Conversational AI":
+    elif project_type == enums.SampleProjectType.CONVERSATIONAL_AI.value:
         file_name = "sample_projects/conversational_ai.zip"
     else:
         raise Exception("Unknown sample project")
```
SimonDegrafKern: This is a valid solution. Another way to approach this would be to use a "switch case" construct, which would be more natural for me when evaluating multiple conditions of which only one can be true. Python had no out-of-the-box switch-case construct for a long time and it got first introduced in python 3.10 if I remember correctly, so it could be that it would not even work for us to make use of the construct because I am not sure if we have a python version > 3.10 everywhere, but still wanted to mention it here just in case. So maybe no real todo for this comment but still good to think about imo

Good example of python switch-case can be found here: https://www.freecodecamp.org/news/python-switch-statement-switch-case-example/

SimonDegrafKern: Maybe it could make sense to include the given project_type in the exception message to give more information for a dev in case of an error and make bug tracking easier. I think there is all the time a conflict between reducing the information displayed to a minimum but still supporting the maintainability of an application as much as possible. But I am in support of including at least the essential information about why an error occurred into the message

controller/transfer/project_transfer_manager.py
```diff
@@ -148,11 +149,12 @@ def import_file(
     """
     send_progress_update_throttle(project_id, task_id, 0)
     project_item = project.get(project_id)
-    project_item.name = data.get(
-        "project_details_data",
-    ).get(
-        "name",
-    )
+    if not project_item.name:
```
SimonDegrafKern, line 152: In what case would this evaluate to true? (ofc if there is no project name, but I am atm a bit confused when this would be the case)

SimonDegrafKern: I just saw as soon as I went into the next file, that in the mutation the name is now optional, I was not aware of this

controller/transfer/project_transfer_manager.py
lumburovskalina: Yes, the switch case would be more natural, I tried that but it says exactly as you say. 

Screenshot 2023-07-19 at 11 19 20

lumburovskalina: yes, it would make sense to include it

controller/transfer/project_transfer_manager.py
lumburovskalina: Okay, then this comment can be resolved?

SimonDegrafKern: yes :)

SimonDegrafKern (approved):

SimonDegrafKern:
merge conflict for submodules model 

- [x] resolved

SimonDegrafKern (approved):

PR #142, opened 2023-07-20
Qdrant filter integrations

lumburovskalina:
PRs:
- https://github.com/code-kern-ai/refinery-ui/pull/150
- https://github.com/code-kern-ai/refinery-gateway/pull/142
- https://github.com/code-kern-ai/refinery-submodule-model/pull/49
- https://github.com/code-kern-ai/refinery-neural-search/pull/31
- https://github.com/code-kern-ai/refinery-embedder/pull/52


controller/record/neural_search_connector.py
```diff
@@ -1,16 +1,20 @@
...
-    project_id: str, embedding_id: str, record_id: str, limit: int
+    project_id: str,
+    embedding_id: str,
+    record_id: str,
+    limit: int,
+    att_filter: Optional[List[Dict[str, Any]]] = None,
 ) -> List[str]:
     url = f"{BASE_URI}/most_similar?project_id={project_id}&embedding_id={embedding_id}&record_id={record_id}&limit={limit}"
 
     # changed from get to post so we can send the filter -> however currently filter isn't part of the prototype so None
```
SimonDegrafKern: could it be that this comment should be removed since it is not None anymore?

SimonDegrafKern:
I found something which I am not sure why it happens, but I try to describe it as best as I can:

Context: 
- I am using the Open AI ada Model
- I am using a 100 records dataset which consists of the first 100 records of the clickbait dataset
[clickbait - initial_default_100 records.json.zip](https://github.com/code-kern-ai/refinery-gateway/files/12319398/clickbait.-.initial_default_100.records.json.zip)
- The embedding I use is on the headline attribute and running_id is the selected filter attribute
- I am running the neural-search, gateway with submodule model and ui in excluded containers which are set on the feature branch

Now the situation: 
If I search for records between 1 and 100 on similarity for the record with running_id 1,
image
than I get 6 records as result which all have a running_id which is smaller than 50. 
image
Here you can't see all of them but I checked multiple times and it is record with running_id [1, 2, 10, 23, 24, 28]

If I now search for records between 1 and 50 again on similarity for record with running_id 1 I get the same result like expected. Again the 6 records [1, 2, 10, 23, 24, 28]. 

If I now search for records between 50 and 100 again on similarity for record with running_id 1 I get a strange result. I get the record with the running_id 1. 

I don't know why this happens tbh. I think, that there is no result in similarity search because of the other tests I did and there should be no match in between running_id 50 and 100. But still I get one record back with running_id 1. Do you know why this happens? Is this the expected behavior from your side? 

- [x] discussed

controller/embedding/manager.py
```diff
@@ -212,3 +213,17 @@ def __recreate_embedding(project_id: str, embedding_id: str) -> Embedding:
...
     return new_embedding_item
+
+
+def update_embedding_payload(
+    project_id: str, embedding_id: str, filter_attributes: List[str]
+) -> None:
+    notification.send_organization_update(
+        project_id=project_id,
+        message=f"upload_embedding_payload:{str(embedding_id)}",
+    )
```
JWittmeyer: Usually we use notifications to send a "finished" note. here it's done before the payload has been updated. So if someone is listening to that value we will run in a racing condition. Sending it last instead should do the trick.

lumburovskalina: My idea here was to send a notification when the update starts (so i can add the spinner on the front-end), because once this is finished we sent a notification 'embedding_updated' and the spinner is removed. Maybe a better naming or the approach is wrong?

JWittmeyer: makes sense. maybe change it to `message=f"upload_embedding_payload:{str(embedding_id):start}"` to prevent irritation

JWittmeyer (approved):

PR #146, opened 2023-08-07
Adds replace or null special char in record data upload

SimonDegrafKern:
PR to fix gateway: https://github.com/code-kern-ai/refinery-gateway/pull/146
tokenizer: https://github.com/code-kern-ai/refinery-tokenizer/pull/37

Todo: https://www.notion.so/kern-ai/more-robust-tokenization-process-ea72185be75b453cac945838864d0a56


controller/transfer/util.py
```diff
@@ -83,6 +84,7 @@ def convert_to_record_dict(
             )
             raise Exception("Upload conversion error", "Upload ran into errors")
         # ensure useable columns dont break the import
+        df = df.replace("\u0000", " ", regex=True)
```
lumburovskalina, line 87: Just as suggestion for the UI:
If we add an information extraction on the attribute, the space will be shown a one token and an be labeled, which maybe confusing because the other spaces are not labelable (if this word exists :D ). Maybe we can add some 'Warning' or 'Note' to inform the user where did the space come from.

lumburovskalina:
If we have a record with multiple "\u0000" in the attribute, this is treated as one space (same behaviour if the spaces in between or if there aren't). Is this wanted behaviour? 
All maybe another UI improvement would be to add some message if the value of the attribute is consisted only from these characters, so it doesn't look empty on the data browser and the labeling page.

Simon: I think thats okay like it is, this should in general be a really rare edge case because usually this character is not used often. By replacing it with one whitespace we already made sure the dataset works again even if there is the bad character and I don't think that there are bad consequences labeling wise after replacing it when it is treated as one space even if there is multiple times the bad char  :) 

- [X] discussed

controller/transfer/util.py
```diff
@@ -141,6 +145,19 @@ def check_and_convert_category_for_unknown(
         )
 
 
+def check_and_covert_nested_attributes_to_text(df: pd.DataFrame) -> pd.DataFrame:
+    for key in df.columns:
+        if [value for value in pick_sample(df, 10, key) if type(value) == dict]:
+            df[key] = df[key].apply(lambda x: json.dumps(x))
```
lumburovskalina, line 152: out of curiosity, can we check the type using the function `isInstance(value,dict)?`

controller/transfer/util.py
```diff
@@ -141,6 +145,19 @@ def check_and_convert_category_for_unknown(
...
+    return df
+
+
+def pick_sample(df: pd.DataFrame, sample_size: int, key: str) -> pd.DataFrame:
+    if len(df[key]) <= sample_size:
```
lumburovskalina: Suggestion (not sure if it's a valid one :D). :
We can move the `sample_size` parameter as the last one, and set it default to 10, so we don't have to pass it every time we want to use the function (if we want to)

controller/transfer/util.py
SimonDegrafKern: Most probably yes, I wrote a comment for this in the other PR :)

controller/transfer/util.py
SimonDegrafKern: Good suggestion I will include it :)

controller/transfer/util.py
SimonDegrafKern: I will add a notification

SimonDegrafKern: After further discussion we decided to not add additional handling since the replacement is handling enough

controller/transfer/util.py
```diff
@@ -141,6 +145,21 @@ def check_and_convert_category_for_unknown(
         )
 
 
+def check_and_covert_nested_attributes_to_text(df: pd.DataFrame) -> pd.DataFrame:
+    for key in df.columns:
+        if [value for value in pick_sample(df, key) if type(value) == dict]:
```
JWittmeyer: not readable & performance wise not optimal

I'd go with a small function that loops the samples and returns True / False & isinstance since it's faster.

Key differences (most likely not noticeable for the small amount):
- isinstance is faster than type(xxx) == 
- no dummy array is created and garbage collected
- if one is a dict we return faster
- readable/ easy to understand code

SimonDegrafKern: Good and specific recommendation, thanks :) 

resolved by simple function

JWittmeyer (approved):

PR #150, opened 2023-09-04
Option to edit records

JWittmeyer:
- https://github.com/code-kern-ai/refinery-ui/pull/155
- https://github.com/code-kern-ai/refinery-gateway/pull/150
- https://github.com/code-kern-ai/refinery-embedder/pull/60

best with `bash start -b edit-option`

caution for Mac Users, refinery-embedder is part of this PR which can result in some performance issues with the branch build
maybe then `bash start -b edit-option -e refinery-embedder` + embedder by hand


controller/embedding/connector.py
```diff
@@ -28,3 +30,31 @@ def request_deleting_embedding(project_id: str, embedding_id: str) -> Any:
...
+    project_id: str, changes: Dict[str, List[Dict[str, Union[str, int]]]]
+) -> None:
+    # example changes structure:
+    # {"<embedding_id>":[{"record_id":"<record_id>","attribute_name":"<attribute_name>","sub_key":"<sub_key>"}]}
+    # note that sub_key is optional and only for embedding lists relevant
+    url = f"{BASE_URI}/re_embed_records/{project_id}"
+    service_requests.post_call_or_raise(url, {"changes": changes})
+
+
+MODEL_PROVIDER_BASE_URI = os.getenv("MODEL_PROVIDER")
```
FelixKirschKern: I think this i not used anywhere

controller/record/manager.py
```diff
@@ -113,3 +124,173 @@ def __reupload_embeddings(project_id: str) -> None:
...
+
+        emb_path = os.path.join(
+            "/inference", project_id, f"embedder-{str(embedding_item.id)}.pkl"
+        )
+        if not os.path.exists(emb_path):
+            errors_found.append(
+                f"can't find embedding PCA for {embedding_item.name}. Try rebuilding or removing the embeddings on settings page."
+            )
+            continue
+        useable_embeddings[embedding_item.attribute_id] = embedding_item
```
FelixKirschKern: An attribute can be embedded multiple times. So instead of this data structure
`useable_embeddings = {}  # dict of UUID(attribute_id): embedding_item`
Useable embeddings need to hold a list of embedding_items per attribute_id.
Currently, we are only updating one of the embeddings (the one that happens to be last in the loop).

And then also change the construction of the embedding_rebuilds list accordingly.

FelixKirschKern:
Found another bug while testing.
When (re)creating a collection in qdrant that contains a filter_attribute, we currently run into an error, because a sql query is broken.
Should be fixed with this PR
https://github.com/code-kern-ai/refinery-neural-search/pull/38

controller/record/manager.py
JWittmeyer: changed

FelixKirschKern (approved):

PR #157, opened 2023-09-21
Changes for cognition

JWittmeyer:
- https://github.com/code-kern-ai/refinery-gateway/pull/157
- https://github.com/code-kern-ai/refinery-submodule-model/pull/56
- https://github.com/code-kern-ai/gates-gateway/pull/37

PRs for cognition launch. Cognition itself runs in dev branch. 
For dev `bash start -x -g -b cognition-init -e cognition-ui,cognition-gateway,refinery-gateway`

starting without exclusion seems to have some issues


alembic/versions/c1fbc725946a_adds_cognition_project_table.py
```diff
@@ -0,0 +1,784 @@
...
+        sa.Column("name", sa.String(), nullable=True),
+        sa.Column("description", sa.String(), nullable=True),
+        sa.Column("color", sa.String(), nullable=True),
+        sa.Column("created_at", sa.DateTime(), nullable=True),
+        sa.Column("operator_routing_source_code", sa.String(), nullable=True),
+        sa.Column("wizard_running", sa.Boolean(), nullable=True),
+        sa.Column(
+            "refinery_synchronization_interval_option", sa.String(), nullable=True
+        ),
+        sa.ForeignKeyConstraint(["created_by"], ["user.id"], ondelete="CASCADE"),
```
FelixKirschKern: If I understand cognition correctly, different users can work on the same project. It could be the case that a user is deleted because he leaves the company, but his colleagues want to keep working on the project.

I think the `ondelete="CASCADE"` is problematic, a better strategy could be to use `ondelete='SET NULL'`.

The same could be a problem in refinery, as we are also using ondelete CASCADE here for project and user.

alembic/versions/c1fbc725946a_adds_cognition_project_table.py
```diff
@@ -0,0 +1,784 @@
...
+        ),
+        sa.ForeignKeyConstraint(
+            ["refinery_question_project_id"], ["project.id"], ondelete="CASCADE"
+        ),
+        sa.ForeignKeyConstraint(
+            ["refinery_references_project_id"], ["project.id"], ondelete="CASCADE"
+        ),
+        sa.ForeignKeyConstraint(
+            ["refinery_relevance_project_id"], ["project.id"], ondelete="CASCADE"
+        ),
```
FelixKirschKern: also not too sure about these `ondelete="CASCADE"`.
As user I wouldn't expect these consequences when deleting a refinery project.

alembic/versions/c1fbc725946a_adds_cognition_project_table.py
```diff
@@ -0,0 +1,784 @@
...
+        ["refinery_relevance_project_id"],
+        unique=False,
+        schema="cognition",
+    )
+
+    op.create_table(
+        "conversation",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("project_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("created_by", postgresql.UUID(as_uuid=True), nullable=True),
```
FelixKirschKern: Here, we don't have a FK Constraint at all

alembic/versions/c1fbc725946a_adds_cognition_project_table.py
```diff
@@ -0,0 +1,784 @@
...
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("project_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("strategy_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("created_by", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("created_at", sa.DateTime(), nullable=True),
+        sa.Column("name", sa.String(), nullable=True),
+        sa.Column("description", sa.String(), nullable=True),
+        sa.Column("strategy_step_type", sa.String(), nullable=True),
+        sa.Column("strategy_step_position", sa.Integer(), nullable=True),
+        sa.ForeignKeyConstraint(["created_by"], ["user.id"], ondelete="CASCADE"),
```
FelixKirschKern: Here again, I think `ondelete="CASCADE"` is what we want as behavior.

FelixKirschKern: There are some similar issues in most of the tables. I think it would make sense to come up with a clear strategy for how we handle ForeignKeys and then go over the model definitions again.

alembic/versions/c1fbc725946a_adds_cognition_project_table.py
```diff
@@ -0,0 +1,784 @@
...
+    op.create_table(
+        "pipeline_logs",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("project_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("message_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("created_at", sa.DateTime(), nullable=True),
+        sa.Column("pipeline_step_type", sa.String(), nullable=True),
+        sa.Column("strategy_step_type", sa.String(), nullable=True),
+        sa.Column("strategy_step_id", postgresql.UUID(as_uuid=True), nullable=True),
+        sa.Column("created_by", postgresql.UUID(as_uuid=True), nullable=True),
```
FelixKirschKern: Logs are created by the pipeline rather than by the user. So on first look, I don't get this column

alembic/versions/c1fbc725946a_adds_cognition_project_table.py
```diff
@@ -0,0 +1,784 @@
...
+        "pipeline_logs",
+        ["strategy_step_id"],
+        unique=False,
+        schema="cognition",
+    )
+
+    op.create_table(
+        "markdown_file",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("user_id", postgresql.UUID(as_uuid=True), nullable=True),
```
FelixKirschKern: Why do we use `user_id` instead of `created_by`?

api/transfer.py
```diff
@@ -187,6 +193,47 @@ async def post(self, request) -> JSONResponse:
...
+    def put(self, request) -> PlainTextResponse:
+        project_id = request.path_params["project_id"]
+        task_id = request.path_params["task_id"]
+        task = upload_task_manager.get_upload_task(
+            task_id=task_id,
+            project_id=project_id,
+        )
+        if task.upload_type != enums.UploadTypes.COGNITION.value:
+            return PlainTextResponse("OK")
+        task.upload_type = enums.UploadTypes.DEFAULT.value
```
FelixKirschKern, line 205: I don't understand, why we need to set the `upload_type` to default, tried to have a look in, but it's really hard to get the underlying logic

controller/embedding/manager.py
```diff
@@ -66,11 +88,17 @@ def delete_embedding(project_id: str, embedding_id: str) -> None:
...
 def __embed_one_by_one_helper(project_id: str, embeddings_ids: List[str]) -> None:
+    ctx_token = general.get_ctx_token()
     for embedding_id in embeddings_ids:
         connector.request_embedding(project_id, embedding_id)
         time.sleep(5)
+        c = 1
         while util.has_encoder_running(project_id):
+            c += 1
+            if c > 12:
+                ctx_token = general.remove_and_refresh_session(ctx_token, True)
```
FelixKirschKern, line 99: i think here is a `c=1` missing, otherwise we would remove and refresh every five seconds after a minute

controller/gates/manager.py
```diff
@@ -0,0 +1,31 @@
...
+
+from . import gates_service
+
+
+def start_gates_container(
+    project_id: str, select_available_heuristics: bool = True
+) -> bool:
+    heuristics, embeddings = [], []
+    if select_available_heuristics:
+        heuristics, embeddings = __get_relevant_gate_ids(project_id)
```
FelixKirschKern: typo `__get_relevant_gates_ids`, `s` in `gates` missing

controller/task_queue/handler/task_queue.py
```diff
@@ -0,0 +1,105 @@
...
+    project_id = task["project_id"]
+    if not project_db_bo.get(project_id):
+        # parent project was deleted so the queue item is done
+        return True
+    sub_task_id = task.get("sub_task_id")
+    if not sub_task_id:
+        # no sub task yet
+        __start_sub_task(task)
+        return False
+    if sub_task_id != TaskType.TASK_QUEUE_ACTION.value:
```
FelixKirschKern, line 43: does the `sub_task_id` holds the `TaskType`, I am a bit confused by the condition (just by the names, haven't digged deeper)

controller/task_queue/manager.py
```diff
@@ -62,14 +85,66 @@ def get_task_function_by_type(task_type: str) -> Tuple[Callable, Callable, int]:
...
+        queue = task_queue.get_task_queue()
+    queue.add_task(task, start_func, check_func, check_every)
 
 
 def remove_task_from_queue(project_id: str, task_id: str) -> None:
     task_queue_db_bo.remove_task_from_queue(project_id, task_id, with_commit=True)
     # no need to dequeue from tasks since the initial check should handle it
+
+
+def __execute_action(project_id: str, user_id: str, action: Dict[str, Any]):
```
FelixKirschKern: typing -> None

controller/task_queue/manager.py
```diff
@@ -62,14 +85,66 @@ def get_task_function_by_type(task_type: str) -> Tuple[Callable, Callable, int]:
...
+        if not embedding_item:
+            raise ValueError(f"Unknown embedding {embedding_name}")
+
+        data_slice_manager.create_outlier_slice(
+            project_id, user_id, str(embedding_item.id)
+        )
+    elif action_type == enums.TaskQueueAction.START_GATES.value:
+        project_id = action.get("project_id")
+        if not project_id:
+            raise ValueError("Missing project id")
```
FelixKirschKern, line 125: A task of type task queue is normally related to one project, but we could theoretically also start gates containers for other projects. Or why do we use the project_id from the action here and not the one provided as argument to the function?

controller/transfer/cognition/constants.py
```diff
@@ -0,0 +1,237 @@
...
+                "name": "Language",
+                "type": DataTypes.CATEGORY.value,
+                "code_build": {"endpoint": "language_detection"},
+            },
+            {
+                "name": "reference_chunks",
+                "type": DataTypes.EMBEDDING_LIST.value,
+                "code": REFERENCE_CHUNKS_SPLIT.replace(
+                    "@@target_attribute@@", "reference"
+                ),
```
FelixKirschKern: could we also use the bricks endpoint now (as it is released),
then with `"code_build": {"endpoint": "newline_splitter"}`

controller/transfer/project_transfer_manager.py
```diff
@@ -913,7 +913,12 @@ def __post_processing_import_threaded(
     project_id: str, task_id: str, embedding_ids: List[str], data: Dict[str, Any], user_id: str
 ) -> None:
     time.sleep(5)
+    ctx_token = general.get_ctx_token()
+    c = 1
     while True:
+        c+=1
+        if c >12:
+            ctx_token = general.remove_and_refresh_session(ctx_token,True)
```
FelixKirschKern: refreshed every time after c > 12

controller/transfer/cognition/import_preparator.py
```diff
@@ -0,0 +1,58 @@
...
+from ..util import read_file_to_df
+from util import category
+
+
+def prepare_cognition_import(project_id: str, task: UploadTask) -> None:
+    # pre init to ensure we can always append an error
+    file_additional_info = __get_blank_file_additional_info()
+    project_item = project.get(project_id)
+    if not project_item:
+        file_additional_info["errors"].append("Can't find project".format(e))
```
FelixKirschKern: format missing a placeholder for `e`

controller/transfer/cognition/import_wizard.py
```diff
@@ -0,0 +1,628 @@
...
+        }
+    )
+
+    # wait for initial tokenization to finish
+    c = 0
+    while True:
+        time.sleep(1)
+        c += 1
+        if c > 120:
+            ctx_token = general.remove_and_refresh_session(ctx_token, True)
```
FelixKirschKern, line 185: refreshed every time after c > 120

controller/transfer/cognition/import_wizard.py
```diff
@@ -0,0 +1,628 @@
...
+    general.commit()
+    general.remove_and_refresh_session(ctx_token, False)
+    notification.send_organization_update(
+        cognition_project_id,
+        "cognition_prep:state:DONE",
+        organization_id=organization_id,
+    )
+
+
+def __add_websocket_message_queue_item(
```
FelixKirschKern, line 244: This method ist not used atm, should we keep it for future usage?

controller/transfer/cognition/import_wizard.py
```diff
@@ -0,0 +1,628 @@
...
+            language_key=None,
+            target_data={"target_type": "lf"},
+        )["person_extraction"]["code"],
+        flush=True,
+    )
+    # from .wizard_function_templates import REFERENCE_CHUNKS_SPLIT
+
+    # print(
+    #     REFERENCE_CHUNKS_SPLIT.replace("@@target_attribute@@", "reference"), flush=True
+    # )
```
FelixKirschKern: Reminder to clean up before merging

graphql_api/query/misc.py
```diff
@@ -60,7 +59,7 @@ def resolve_has_updates(self, info) -> bool:
         return manager.has_updates()
 
     def resolve_helper(self, info) -> bool:
-        import controller.transfer.export_parser as export_parser
+        import controller.transfer.cognition.import_wizard as w
 
-        export_parser.query_builder_dummy()
+        w.dummy()
```
FelixKirschKern, line 64: Reminder to clean up before merging

controller/task_queue/manager.py
JWittmeyer: long story short -> yes. This is because there is one task_queue for the whole wizard thing (to get an overall progress). So to overwrite it we can provide an id

controller/transfer/cognition/import_wizard.py
JWittmeyer: yes

FelixKirschKern (approved):

PR #160, opened 2023-10-02
Weak Supervision without manual labels

FelixKirschKern:
Main PR:
https://github.com/code-kern-ai/refinery-weak-supervisor/pull/33

### DIFF
--- a/graphql_api/mutation/weak_supervisor.py
+++ b/graphql_api/mutation/weak_supervisor.py
@@ -159,11 +159,6 @@ class RunInformationSourceAndInitiateWeakSupervisionByLabelingTaskId(graphene.Mu
         pl_manager.create_payload(
             project_id, information_source_id, user.id, asynchronous=False
         )
-        ws_overwrite = None
-        if overwrite_weak_supervision is not None:
-            ws_overwrite = overwrite_weak_supervision
-        else:
-            ws_overwrite = overwrite_default_precision
 
         source_names = []
         labeling_task_item = labeling_task.get(project_id, labeling_task_id)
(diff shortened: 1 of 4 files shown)


graphql_api/mutation/weak_supervisor.py
```diff
@@ -116,29 +140,46 @@ class Arguments:
...
-                source_statistic.true_positives > 0
-                for source_statistic in information_source.source_statistics
-                if source_statistic.true_positives is not None
+            information_source.is_selected = (
+                any(
+                    source_statistic.true_positives > 0
+                    for source_statistic in information_source.source_statistics
+                    if source_statistic.true_positives is not None
+                )
+                or ws_overwrite is not None
```
JWittmeyer: maybe i'm not understanding it right but would this result in all being selected if either of the overwrite parameters is set? So e.g. a function that didn't run yet would still be concidered selected.

graphql_api/mutation/weak_supervisor.py
```diff
@@ -116,29 +140,46 @@ class Arguments:
...
             project_id, information_source_id, user.id, asynchronous=False
         )
+        ws_overwrite = None
+        if overwrite_weak_supervision is not None:
+            ws_overwrite = overwrite_weak_supervision
+        else:
+            ws_overwrite = overwrite_default_precision
+
         labeling_task_item = labeling_task.get(project_id, labeling_task_id)
         for information_source in labeling_task_item.information_sources:
```
JWittmeyer, line 168: I always thought Run+WS only selects the current function (which was misinformation on my side) but it selects all from the task which doesn't match the description in our docs (which all all)
image

graphql_api/mutation/weak_supervisor.py
FelixKirschKern: changed so that only finished are selected, corresponding to UI behavior on weak supervision page

graphql_api/mutation/weak_supervisor.py
FelixKirschKern: will inform devrels for documentation update

JWittmeyer (approved):

JWittmeyer (approved):

PR #196, opened 2024-03-12
FastAPI Integration

anmarhindi:
Main PR

Related PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/196
- https://github.com/code-kern-ai/refinery-ui/pull/15
- https://github.com/code-kern-ai/refinery-submodule-model/pull/86
- https://github.com/code-kern-ai/submodule-javascript-functions/pull/9
- https://github.com/code-kern-ai/dev-setup/pull/35
- https://github.com/code-kern-ai/deployment-dev-app/pull/19
- https://github.com/code-kern-ai/refinery-updater/pull/61

Start with `bash start -b fastapi -g -e refinery-ui,refinery-gateway,refinery-updater`


fast_api/routes/attribute.py
```diff
@@ -0,0 +1,18 @@
+from controller.attribute import manager
+from typing import List, Union
+from fast_api.routes.client_response import pack_json_result
+from fastapi import APIRouter, Query
+from submodules.model.util import sql_alchemy_to_dict
+
+router = APIRouter()
+
+
+@router.get("/all-attributes/{project_id}")
```
anmarhindi: Should resolve to:
`@router.get("/{project_id}/attributes_by_project_id")`

anmarhindi: In the middleware, we are checking for the project id in the specified order for access.

middleware/database_session.py
```diff
@@ -3,13 +3,18 @@
...
 from middleware.query_mapping import path_query_map
-from route_prefix import PREFIX_PROJECT
+from route_prefix import PREFIX_ATTRIBUTE, PREFIX_PROJECT
 from submodules.model.business_objects import general
 from controller.auth import manager as auth_manager
 
 logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger(__name__)
 
+project_access_prefix = [
```
JWittmeyer: also capitalized as also a const

middleware/database_session.py
```diff
@@ -37,13 +42,15 @@ async def dispatch(self, request, call_next):
...
-                project_id = url_split[0]
-                most_likely_url_part = url_split[1]
-                if project_id:
-                    auth_manager.check_project_access(info, project_id)
+            for prefix in project_access_prefix:
+                if prefix in request.url.path:
+                    most_likely_url_part: str = request.url.path.replace(prefix, "")
+                    url_split = most_likely_url_part.split("/")
+                    project_id = url_split[0]
+                    most_likely_url_part = url_split[1]
```
JWittmeyer: the most_likely_url_part var isn't used afterawrds

middleware/database_session.py
```diff
@@ -12,13 +16,52 @@ def __init__(self, app):
         super().__init__(app)
 
     async def dispatch(self, request, call_next):
+        info = self._prepare_info(request)
 
-        session_token = general.get_ctx_token()
+        request.state.session_token = general.get_ctx_token()
```
JWittmeyer, line 25: we dont need to set a session token for every request but rather for every task or thread.

In this instance this means that we should only remove the token in the _check_access function if an exception occured and we need to send and unauthorized response + remove line 26 (second token request)

lumburovskalina:
The PR for the deployment-dev needs to be open with the changes from dev-setup
- [x] resolved

lumburovskalina:
Note sure if part of the task but there are some queries and mutations on the admin dashboard (some are the same as in refinery - createOrganization, addUserToOrg, operations with comments..)
- [ ] resolved
- [x] ignored

fast_api/routes/data_slices.py
```diff
@@ -0,0 +1,76 @@
...
+from util import notification
+import json
+
+
+router = APIRouter()
+
+
+@router.get(
+    "/{project_id}", dependencies=[Depends(auth_manager.check_project_access_dep)]
+)
```
lumburovskalina, line 20: I think it makes to add /projectId/data-slices for getting all dataSlices, since this request is used in many pages and it would be much easier to find it with 'name' than with id.(not crutual for all requests but i think for this one makes sense).
- [x] resolved
- [ ] ignored

lumburovskalina:
Screenshot 2024-04-09 at 16 27 27

The zeroshot models are not sorted correctly, they should be sorted by size
- [ ] resolved
- [x] ignored


Anmar: will ignore for now as not directly fastapi related

lumburovskalina:
Screenshot 2024-04-09 at 16 30 32

The query for notifications does not work. Can be reproduced by clicking on the notifications icon on the header or by clicking on the notifications that pop up (not notificationsByUserId, but the query for getting all the notifications)
- [x] resolved

fast_api/routes/data_slices.py
anmarhindi: renamed to get-data-slices for clarity and searchability

anmarhindi:
Screenshot 2024-04-09 at 16 38 11

Issues with weak-supervision-run
- [x] resolved

lumburovskalina:
Screenshot 2024-04-09 at 17 09 11

Error when getting the model callbacks, JSON.parse should be removed
- [x] resolved

lumburovskalina:
On fastapi branch: 
Screenshot 2024-04-09 at 17 20 40
On dev: 
Screenshot 2024-04-09 at 17 20 05

For the query queuedTasks, the name in the table is missing
- [x] resolved

lumburovskalina:
Screenshot 2024-04-09 at 17 27 11

The statistics are missing on the heuristics page
- [x] resolved

lumburovskalina:
Screenshot 2024-04-09 at 17 32 48

The query recordByRecordId is this in graphql
- [x] resolved

lumburovskalina:
Screenshot 2024-04-09 at 17 44 23

Zeroshot - run on 10 records has error -> can be ignored for now since it's not part of this PR, there is some issue when a project with zero shot is imported, it fails because the input attribute is not set correctly
- [ ] resolved
- [x] ignored

lumburovskalina:
Payload by payload id doesn't give the right iteration
- [x] resolved

lumburovskalina:
uniqueValuesByAttributes needs to be stringified in the BE and parsed in the FE, my mistake :) 
- [x] resolved

lumburovskalina:
Some issues with slices: 
When i select a static slice the application becomes really slow, you can barely change the filter, then I get a notification that the filter update didn't work
When i select a dynamic slice the values of the slice are not applied when searching in the databrowser
- [x] resolved

lumburovskalina:
Updating slices by name doesn't work
- [x] resolved

lumburovskalina:
Screenshot 2024-04-09 at 18 16 52

The query: prepareProjectExport is still in graphql
- [x] resolved

lumburovskalina:
Screenshot 2024-04-10 at 09 40 08

Link-locked doesn't work when I try to access the labeling page as an annotator or an expert
- [x] resolved

lumburovskalina:
Embeddings does not store filterAttributes even though they are sent from the front-end correctly. With this, the update embedding payload doesn't work as expected. Also the similarity search with filter attributes does not work. 
Are the filterAttributes part of the whitelist for embedding?

_________________
jw edit:
The request format changed for the attributes field so the value exists in the database but the frontend doesn't parse it correctly (same on the databrowser imo)

- [x] resolved

lumburovskalina:
Screenshot 2024-04-10 at 14 35 06

RecordIDE works but returns the result in a really strange way. Then the next request fails
- [x] resolved

fast_api/routes/heuristic.py
```diff
@@ -0,0 +1,295 @@
...
+    manager.toggle_information_source(project_id, information_source_id)
+    notification.send_organization_update(
+        project_id, f"information_source_updated:{information_source_id}"
+    )
+    return pack_json_result({"data": {"toggleInformationSource": {"ok": True}}})
+
+
+@router.post(
+    "/{project_id}", dependencies=[Depends(auth_manager.check_project_access_dep)]
+)
```
lumburovskalina, line 169: i think this should be renamed, the route has only projectId but what it actually does is selecting or deselecting all heuristics

lumburovskalina:
Update knowlegde base/lookup list does not work. I tried to update the name  (however, updating the description works).
- [x] resolved

lumburovskalina:
Paste terms to lookuplists/knowlegde bases does not have error, but doesn't add the terms to the list (i think the data is not sent correctly)
- [x] resolved

lumburovskalina:
Screenshot 2024-04-10 at 17 16 31

The results needs to be wrapped in a token because the front end expects it like this : res['data']['createPersonalAccessToken']['token'] -> my mistake :) 
___________________
jw edit: Removed the ['token'] from frontend



- [x] resolved

lumburovskalina:
Delete label doesn't work (no error in the request but still does not delete the label)
- [x] resolved

lumburovskalina:
Screenshot 2024-04-10 at 17 40 56

Create tasks and labels in the bricks integrator does not work

_______________________________________
jw edit: Also on dev => fixed with fake debounce


- [x] resolved

JWittmeyer:
Model provider info is recollected on every change of the embedding text
image

- [x] resolved

controller/embedding/manager.py
```diff
@@ -137,6 +145,63 @@ def get_embedding_name(
...
+    "type",
+    "state",
+    "progress",
+    "dimension",
+    "count",
+    "platform",
+    "model",
+    "filter_attributes",
+    "attribute_id",
+    "onQdrant",
```
lumburovskalina: why onQdrant is written with camelcase?

fast_api/routes/heuristic.py
JWittmeyer: added `/change-selection-state`

controller/embedding/manager.py
JWittmeyer: falsely added here, not part of the table columns but an added parameter outside the table => removed onQdrant from whitelist

controller/project/manager.py
```diff
@@ -34,12 +35,44 @@
 from controller.transfer.manager import check_and_add_running_id
 from controller.upload_task import manager as upload_task_manager
 from controller.gates import gates_service
+from controller.auth import kratos
+from submodules.model.util import sql_alchemy_to_dict
+
+LABELING_TASK_WHITELIST = {
+    "id",
+    "name",
+    "taskTarget",
```
lumburovskalina: camel case, i think this is wrong?
below the same for attribute whitelist and information source whitelist

lumburovskalina:
Empty init.py file in routes. 
- [ ] resolved
- [x] ignored
________
jw edit:
they are required for pyhton folders
image

controller/project/manager.py
JWittmeyer: removed unused WHITELISTS

fast_api/routes/attribute.py
```diff
@@ -0,0 +1,92 @@
...
+                "calculateUserAttributeSampleRecords": {
+                    "record_ids": record_ids,
+                    "calculated_attributes": calculated_attributes,
+                }
+            }
+        }
+    )
+
+
+@router.delete("/{project_id}/delete-user-attribute")
```
lumburovskalina: dependencies for project_id missing

fast_api/routes/project_setting.py
```diff
@@ -0,0 +1,344 @@
...
+    request: Request,
+    project_id,
+):
+    user_id = auth_manager.get_user_by_info(request.state.info).id
+    project_manager.update_project_for_gates(project_id, user_id)
+    return pack_json_result({"data": {"updateProjectGates": {"ok": True}}})
+
+
+@router.post("/{project_id}/calculate-user-attribute-all-records")
+def calculate_user_attribute_all_records(
```
lumburovskalina, line 259: dependencies missing

fast_api/routes/project_setting.py
```diff
@@ -0,0 +1,344 @@
...
+        },
+        True,
+    )
+
+    return pack_json_result(
+        {"data": {"calculateUserAttributeAllRecords": {"ok": True}}}
+    )
+
+
+@router.post("/{project_id}/create-task-and-labels")
```
lumburovskalina: dependencies missing

lumburovskalina:
Screenshot 2024-04-11 at 15 18 06

The statistics part is visible but the number of records are not displayed and the precision is always n/a.
- [x] resolved

lumburovskalina:
Screenshot 2024-04-11 at 15 26 01

Search without filter doesn't work. The normal Search works
- [x] resolved

lumburovskalina:
Saving data slices doesn't work. When I click on the slice again to see the results it does not give the correct values. (the update slice by name and the regular update works) 
- [x] resolved

lumburovskalina:
Screenshot 2024-04-11 at 16 26 08

Delete project hasnot be replaced on one place: when removing from project list page (only admins can do that)
- [x] resolved

JWittmeyer:
Why are so many endpoints with async?
image

we did experience a lot of issues with blocking async calls in a multi user environment.

- [ ] resolved
- [x] explained why / ensured none blocking

Anmar: The `request.json()` is an asynchronous operation that requires the async declaration. 

__________________________________________
@anmarhindi Not enough unfortunatly. Async blocks until an await is called. so, long running requests will block all others until the async request has an await or is completed. Since our database connection isn't async this will result in issues with multiple user. To test this you can open another tab & run the prepare download endpoint in multiple screens & check the network tab time. 

This can be circumvented (to some extend) by using def endpoints as they are threaded differently and can run simultaneously.


For the prepare download endpoint i changed it in [this commit](https://github.com/code-kern-ai/refinery-gateway/pull/196/commits/eb95d97bf3faa8ff2168e0de5b6617786a9b6179). Now running export in two screens simultaneously "works")

Please change wherever possible (by providing a body in the function signature and removing the async)

- [x] resolved

@JWittmeyer Agreed. I prefer the Body impl. that I have used. Just an FYI the async in this case does not block in a similar way other async code would in the main thread, as these FastAPI endpoints are dealt with concurrently to prevent such a scenario you described. prepare-record-export itself contains multiple network calls that can congest your home network so this would affect any endpoint if there was a big download/upload happening on your device at the same time. However, the implementation of Body is neater, moving forward this will be the preferred approach.

lumburovskalina:
Labeling function - Run on 10 for information extraction does not work
- [x] resolved

lumburovskalina:
Screenshot 2024-04-15 at 10 26 05

I get an error when i try to access the labeling page for the first time, and I cannot add extraction labels.
- [x] resolved

lumburovskalina:
Screenshot 2024-04-15 at 13 41 30
 
I tried to create an outlier, but i got an error for static slice?
- [x] resolved

Lina: i am looking into this now

lumburovskalina:
Screenshot 2024-04-15 at 13 45 05

The filter for heuristic(only different resutl) in a labeling task doesn't work. I added Manual label 'No', it worked, when I added only with different results it didn't work (all the other heuristics work)
- [x] resolved

Lina: i am looking into this now

lumburovskalina (approved):

PR #200, opened 2024-03-28
Flying docs

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-ui/pull/39


controller/transfer/cognition/minioUpload.py
```diff
@@ -0,0 +1,37 @@
...
+from controller.task_queue import manager as task_queue_manager
+
+
+def handle_cognition_file_upload(path_parts: List[str]):
+
+    if path_parts[1] != "_cognition":
+        return
+
+    if path_parts[3] == "chat_tmp_files":
+        if path_parts[5] == "queued":
```
FelixKirschKern: What do you think of combining the conditions as we're not distinguishing between both states further?

controller/transfer/cognition/minioUpload.py
```diff
@@ -0,0 +1,37 @@
...
+                conversation_item.created_by,
+                {
+                    "cognition_project_id": cognition_project_id,
+                    "conversation_id": conversation_id,
+                    "minio_path": "/".join(path_parts[1:]),
+                    "bucket": path_parts[0],
+                },
+                True,  # not sure if prio is right here as the prio tasks should only take < 1 min but waiting for the normal queue will take ages depending on the queue
+            )
+        return
```
FelixKirschKern: no need for the return here

controller/transfer/cognition/minioUpload.py
```diff
@@ -0,0 +1,37 @@
+from typing import List
```
FelixKirschKern: file name is in camelCase

controller/transfer/cognition/minioUpload.py
JWittmeyer: was initially planned with multiple stages, not the case anymore => combined

FelixKirschKern (approved):

PR #209, opened 2024-04-24
FastAPI Integration

anmarhindi:
Link to Main PR:
- https://github.com/code-kern-ai/cognition-gateway/pull/66


fast_api/types.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+class ExtendedSearch:
+    def __init__(
+        self,
+        sql: str = None,
+        query_limit: int = None,
+        query_offset: int = None,
+        full_count: int = None,
+        session_id: UUID = None,
```
JWittmeyer, line 35: we seldom use UUID for ids. To double check => is this here intended?

JWittmeyer:
expected a requirements change for the removed packages for gql & correct version from fastapi & starlette

Note that currently the starlette & fastapi are commented out in the common-requirements.txt & old versions (to support gql via starlette) were added in the requirements.in. Reach out if we should do a quick intro to this :) 

- [x] resolved

fast_api/types.py
anmarhindi: Yes, in a specific case, we have UserSessions returned and assigning its id to session_id in calls such as `resolve_records_by_static_slice` and `resolve_extended_search`. UserSessions is returning python UUIDs via the param as_uuid=True

fast_api/routes/data_browser.py
```diff
@@ -258,3 +292,18 @@ def update_data_slice(
         handle_error(e, user.id, project_id)
 
     return pack_json_result({"data": {"updateDataSlice": {"ok": ok}}})
+
+
+def handle_error(exception: Exception, user_id: str, project_id: str):
+    general.rollback()
+    if exception.__class__ == TooManyRecordsForStaticSliceException.__class__:
```
JWittmeyer: maybe best with instance of as the class string compare feels off
```suggestion
    if isinstance(exception,TooManyRecordsForStaticSliceException):
```

fast_api/routes/organization.py
```diff
@@ -18,6 +27,7 @@
 from fast_api.routes.client_response import pack_json_result
 from submodules.model import events
 from submodules.model.business_objects import organization
+from submodules.model.business_objects.user import get
```
JWittmeyer: without as <xx> or indicator a "get" is hard to identify where it's coming from. Usually we import the full object (also to prevent parial import issues) so
```suggestion
from submodules.model.business_objects import user as user_db_bo
```

so we can go with `user_db_bo.get(xxx)` and already see whats happening. 

For some clearly described functions we also work with function imports but since all db objects have a get not optimal here

fast_api/routes/organization.py
```diff
@@ -56,11 +66,47 @@ def get_user_info(request: Request):
...
+        }
+    }
+
+    return pack_json_result({"data": data})
+
+
+@router.get("/get-user-info-mini")
+def get_user_info_mini(request: Request):
+    user_id = str(auth_manager.get_user_by_info(request.state.info).id)
+    user = get(user_id)
```
JWittmeyer: we first get the user to extract the id to then get the user?
(the object returned by get_user_by_info is the one from the db already)

```suggestion
    user = auth_manager.get_user_by_info(request.state.info)
    user_id = str(user.id)
```

same for above

middleware/database_session.py
```diff
@@ -1,143 +1,141 @@
...
-        finally:
-            if log_request:
-                # after call next so the path_params are mapped
-                await self._log_request(request)
-            general.remove_and_refresh_session(request.state.session_token)
-        return response
+async def handle_db_session(request: Request, call_next):
+    info = _prepare_info(request)
+    # await set_body(request)
+    await request.body()
```
JWittmeyer: this now sets it on every request which was chosen because of gql (which needed it on every request as all are POST)

For the middleware in fastapi i would orient this towards the ones used in gates / cogntion.

so for setting the body only when necessary:
``` python
    log_request = log_storage.extract_state_info(request, "log_request")
    length = request.headers.get("content-length")
    if log_request and length and int(length) > 0:
        # streams (e.g. json() call) can only be consumed once so we cache it as the mapping from fastapi to our body classes also needs it
        await log_storage.set_request_data(request)
    try:
        response = await call_next(request)
    finally:
        if log_request:
            # after call next so the path_params are mapped
            await log_storage.log_request(request)
        general.remove_and_refresh_session(session_token)

```

middleware/database_session.py
```diff
@@ -1,143 +1,141 @@
...
+    return FastAPIResolveInfo(
+        context={"request": None},
+        field_name=field_name,
+        parent_type=parent_type,
+    )
+
+
+# async def set_body(request: Request):
+#     # await & body access in middleware aren't straightforward https://github.com/tiangolo/fastapi/issues/394
+#     # fixed in newer versions https://github.com/tiangolo/fastapi/discussions/8187#discussioncomment-7962881
```
JWittmeyer: can be removed

middleware/database_session.py
```diff
@@ -1,143 +1,133 @@
...
 
-        return None
+    log_request = auth_manager.extract_state_info(request, "log_request")
+    length = request.headers.get("content-length")
 
-    def _prepare_info(self, request):
-        field_name = None
-        parent_type = None
+    if log_request and length and int(length) > 0:
+        await request.body()
```
JWittmeyer: `await log_storage.set_request_data(request)` (the function used in gates / cognition) should be used and adapted. The await request.body() doesn't do much only awaiting. The value also needs to be stored in a state. 

From gates:

```
async def set_request_data(request: Request) -> bytes:
    data = None
    length = request.headers.get("content-length")
    if length and int(length) > 0:
        if request.headers.get("Content-Type") == "application/json":
            data = await request.json()
        else:
            data = await request.body()
    request.state.data = data

```

and then inside the log_request function the data is used where it's available

```
    data = None
    if hasattr(request.state, "data"):
        data = request.state.data
```

anmarhindi: Initially I thought (from the Github issues via fastapi repo) that the new changes made was to allow the request.body() call to directly inject the data into the request, but now I can see this is not the case. The changes you have suggested have been adopted.

JWittmeyer (approved):

PR #213, opened 2024-05-13
Pagination users

lumburovskalina:
Link to main PR: https://github.com/code-kern-ai/admin-dashboard/pull/26


fast_api/routes/organization.py
```diff
@@ -342,3 +331,48 @@ def archive_admin_message(
...
+        },
+        wrap_for_frontend=False,
+    )
+
+
+@router.delete("/delete-user")
+def delete_user(request: Request, body: DeleteUserBody = Body(...)):
+    auth_manager.check_admin_access(request.state.info)
+    user_manager.delete_user(body.user_id)
+    return pack_json_result({"data": {"deleteUser": {"ok": True}}})
```
LennartSchmidtKern: If this is a new entpoint we do need to mimic the response pattern(date + query name) of graphql because it was never an query nedpoint of it

fast_api/routes/organization.py
```diff
@@ -342,3 +331,48 @@ def archive_admin_message(
...
+    ]
+    active_users = {user["id"]: user for user in active_users}
+
+    data = user_manager.get_mapped_sorted_paginated_users(
+        active_users, body.sort_key, body.sort_direction, body.offset, body.limit
+    )
+    return pack_json_result(
+        {
+            "data": {
+                "mappedSortedPaginatedUsers": data,
```
LennartSchmidtKern: If this is a new entpoint we do need to mimic the response pattern(date + query name) of graphql because it was never an query endpoint of it

fast_api/routes/organization.py
```diff
@@ -342,3 +331,48 @@ def archive_admin_message(
 def set_language_display(request: Request, body: UserLanguageDisplay = Body(...)):
     user_manager.update_user_language_display(body.user_id, body.language_display)
     return pack_json_result({"data": {"changeUserLanguageDisplay": {"ok": True}}})
+
+
+@router.post("/mapped-sorted-paginated-users")
+def get_mapped_sorted_paginated_users(
+    request: Request, body: MappedSortedPaginatedUsers = Body(...)
+):
+    active_users = user_manager.get_active_users(body.filter_minutes, None)
```
LennartSchmidtKern, line 339: Every endpoint we add for the backend, which returns admin sensitive data must also be protected by check admin access as for earlier (simplified) active_users. Otherwise we create an information oracle for normal users with critical information exposed.

controller/user/manager.py
```diff
@@ -92,7 +92,7 @@ def remove_organization_from_user(user_mail: str) -> None:
 
 def get_active_users(minutes: str, order_by_interaction: bool) -> User:
     now = datetime.now()
-    last_interaction_range = (now - timedelta(minutes=minutes)) if minutes else None
+    last_interaction_range = (now - timedelta(minutes=minutes)) if minutes > 0 else None
```
LennartSchmidtKern, line 95: can not compare minutes:str > 0, at runtime it also gets a int, so ther eis no error, but maybe change types

fast_api/routes/organization.py
```diff
@@ -120,8 +124,11 @@ def all_active_admin_messages(request: Request, limit: int = 100) -> str:
 
 @router.get("/all-admin-messages")
 def all_admin_messages(request: Request, limit: int = 100) -> str:
+
```
LennartSchmidtKern: maybe we should also restrict this endpoint for admins, not sure about the purpose

controller/user/manager.py
```diff
@@ -101,3 +101,49 @@ def get_active_users(minutes: str, order_by_interaction: bool) -> User:
...
+            get_user["role"] = active_user_by_id["role"]
+            get_user["organization"] = active_user_by_id["organizationName"]
+
+            final_users.append(get_user)
+
+    final_users = sorted(
+        final_users,
+        key=lambda x: x[sort_key] if sort_key in x else None,
+        reverse=(
+            sort_direction == -1
```
LennartSchmidtKern: could be the error mentioned in the comments: switch 1 and -1 around here

fast_api/routes/organization.py
lumburovskalina: yes, this is the request that displays the admin messages on the admin dashboard (not the one for the users but for us)

controller/user/manager.py
lumburovskalina: no, this is not the reason, this would revert the whole array, the only problem is sorting with numbers

lumburovskalina: changed it in the end :)

LennartSchmidtKern (approved):

PR #214, opened 2024-05-14
Admin Logs

JWittmeyer:
- https://github.com/code-kern-ai/refinery-gateway/pull/214
- https://github.com/code-kern-ai/refinery-submodule-model/pull/100
- https://github.com/code-kern-ai/admin-dashboard/pull/27
- https://github.com/code-kern-ai/cognition-gateway/pull/78
- https://github.com/code-kern-ai/refinery-updater/pull/67
- https://github.com/code-kern-ai/gates-gateway/pull/54

### General Idea
This is a first draft for an admin log approach. The idea is that admin requests are logged in customer organizations. For this a new option in the admin dashboard was introduced.
- DONT_LOG # nothing is logged
- NO_GET  # everything but method GET is logged
- ALL # everything is logged

In most cases NO_GET is the one to go with as this will only log changes in the system. With every request being logs this can be quite the size in a short amount of time.

There are also endpoints that are GET like requests but needed a body options so the method type is POST. For those I added a fast api dependency option to attach to routes with the issue. I didn't check every route so some might still be missing. Worst case is that the logs is a bit bigger than necessary for the NO_GET option. Also gql request are always POST requests but since we are working on removing gql no special handling added for them.

### tasks
- [x] added volumes to dev-setup

(after review)
- [x] added volumes to app dev
- [ ] added volumes to cognition release PR

### best with
`bash start -a -b admin-logs`
or
`bash start -a -b admin-logs -e refinery-gateway,cognition-gateway,gates-gateway`


anmarhindi:
It seems that after starting the application, the welcome screen is blank.
Screenshot 2024-05-15 at 09 22 26

- [x]  resolved

anmarhindi (approved):

PR #232, opened 2024-07-02
Adds customer button table

JWittmeyer:
- https://github.com/code-kern-ai/refinery-gateway/pull/232
- https://github.com/code-kern-ai/refinery-submodule-model/pull/108
- https://github.com/code-kern-ai/submodule-react-components/pull/20
- https://github.com/code-kern-ai/admin-dashboard/pull/40
- https://github.com/code-kern-ai/cognition-ui/pull/75

### Notes

Not planned as more functionality atm since HDI still needs to provide the actual mapping. The whole concept & logic around it can already be tested.

Since in theory the buttons could exist everywhere they are stored in the global db schema. 

I wasn't sure where to best place the endpoints (refinery vs cognition) to reduce PRs i landed on refinery in misc so should be simple to extract / transfer when we change to lite. 

For this change an external mapper was also created. A server already exists, but also works on localhost so clone & start + use url `http://localhost:9060` in config.
https://github.com/code-kern-ai/external-data-mapper
The server is on Digital Ocean & has the base url `https://external.data-mapper.kern.ai`.

Icon strings are from [tabler-icons](https://tabler.io/icons). Use the `Icon<xxx>` notation when adding e.g.
image

Example config with running data mapper on locahost (if the .env in the data-mapper is filled with the corresponding key):
```json
{
    "url": "http://localhost:9060/hdi/map-to-collect-data?key=abc123",
    "icon": "IconAdCircle",
    "tooltip": "Map results to HDI D&O Excel"
}
```

Example config for server (key is faulty but can be asked for):
```json
{
    "url": "https://external.data-mapper.kern.ai/hdi/map-to-collect-data?key=ExyEX4nMBBRab1NijqCpZcrEXz1MQ",
    "icon": "IconExternalLink",
    "tooltip": "External map to HDI Excel"
}
```

best with `bash start -a -b customer-button`

Feature video:
https://drive.google.com/file/d/1fnNh17OCPB3LOIFcCrL7buXocZaTnahx/view?usp=drive_link

______________________________

TODO:
- [x] add some obfuscation logic to custom button keys to prevent that they can be read from the network tab (base64)
- [x] tested firewall on server
- [x] hdi files to server
- [x] set up server
- [x] InfoSec ticket for server


controller/misc/manager.py
```diff
@@ -72,3 +81,113 @@ def __updater_update_to_newest() -> None:
...
+                    raise_me, f"URL {url} answered with {x.status_code}"
+                )
+        except Exception:
+            import traceback
+
+            print(traceback.format_exc(), flush=True)
+            return __raise_or_return(raise_me, f"URL {url} is not reachable")
+
+        return  # returns None so "no error"
+    return __raise_or_return(raise_me, f"Unknown customer button type: {type}")
```
LennartSchmidtKern, line 161: unreachable return

JWittmeyer: intended for us devs as known fallback => wont be changed

LennartSchmidtKern (approved):

PR #241, opened 2024-08-08
Task Master

LennartSchmidtKern:
This is the main PR

Related PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/241
- https://github.com/code-kern-ai/refinery-submodule-model/pull/116
- https://github.com/code-kern-ai/cognition-gateway/pull/121
- https://github.com/code-kern-ai/refinery-tokenizer/pull/68

Deployments:
- https://github.com/code-kern-ai/dev-setup/pull/46
- https://github.com/code-kern-ai/deployment-dev-app/pull/34

Task Master Repo:
- https://github.com/code-kern-ai/cognition-task-master


alembic/versions/af5c30d91685_remove_project_id_task_queue.py
```diff
@@ -0,0 +1,51 @@
...
+    sa.Column('organization_id', postgresql.UUID(as_uuid=True), nullable=True),
+    sa.Column('creation_month', sa.Date(), nullable=True),
+    sa.Column('macro_type', sa.String(), nullable=True),
+    sa.Column('execution_count', sa.Integer(), nullable=True),
+    sa.Column('processed_files_count', sa.Integer(), nullable=True),
+    sa.ForeignKeyConstraint(['organization_id'], ['organization.id'], ondelete='CASCADE'),
+    sa.PrimaryKeyConstraint('id'),
+    sa.UniqueConstraint('organization_id', 'creation_month', 'macro_type', name='unique_macro_summary'),
+    schema='cognition'
+    )
```
JWittmeyer: not sure why this is here and not it's own feature branch

JWittmeyer: (no need to change now)

alembic/versions/af5c30d91685_remove_project_id_task_queue.py
```diff
@@ -0,0 +1,51 @@
...
+    op.create_index(op.f('ix_cognition_macro_execution_summary_creation_month'), 'macro_execution_summary', ['creation_month'], unique=False, schema='cognition')
```
JWittmeyer: looks like it's still in the refinery schema. We wanted to change this to global

JWittmeyer: (which should be more of a drop old table, create new table)

api/transfer.py
```diff
@@ -404,11 +323,14 @@ def init_file_import(task: UploadTask, project_id: str, is_global_update: bool)
         )
     if task.file_type != "knowledge_base":
         only_usable_attributes = task.file_type == "records_add"
-        task_queue_manager.add_task(
-            project_id,
+        project_item = project.get(project_id)
+        org_id = str(project_item.organization_id)
+        task_master_manager.queue_task(
+            str(org_id),
```
JWittmeyer, line 329: already a string so no need to cast it twice

controller/transfer/cognition/import_wizard.py
```diff
@@ -207,9 +213,16 @@ def __finalize_setup(
...
-    task_id, position = task_queue_manager.add_task(
-        reference_project_id, enums.TaskType.TASK_QUEUE, user_id, task_list
+    queue_response = task_master_manager.queue_task(
+        organization_id,
+        user_id,
+        enums.TaskType.TASK_QUEUE,
+        {"project_id": cognition_project_id, "task_list": task_list},
     )
+    queue_info = queue_response.json()
+    print(queue_info, flush=True)
```
JWittmeyer: probably debug

controller/transfer/cognition/import_wizard.py
```diff
@@ -207,9 +213,16 @@ def __finalize_setup(
...
 
-    task_id, position = task_queue_manager.add_task(
-        reference_project_id, enums.TaskType.TASK_QUEUE, user_id, task_list
+    queue_response = task_master_manager.queue_task(
+        organization_id,
+        user_id,
+        enums.TaskType.TASK_QUEUE,
+        {"project_id": cognition_project_id, "task_list": task_list},
     )
+    queue_info = queue_response.json()
```
JWittmeyer: maybe we also should check for response.ok

fast_api/routes/heuristic.py
```diff
@@ -187,24 +187,28 @@ def set_payload(
...
         {
-            "information_source_id": heuristic_id,
+            "project_id": str(project_id),
+            "information_source_id": str(heuristic_id),
             "source_type": information_source_item.type,
         },
         priority=priority,
     )
 
+    queue_id = task_master_response.json().get("task_id")
```
JWittmeyer: what if the response fails?

fast_api/routes/task_execution.py
```diff
@@ -0,0 +1,94 @@
...
+    "/weak-supervision",
+)
+def weak_supervision(
+    weak_supervision_action_execution: WeakSupervisionActionExecutionBody,
+):
+    project_id = weak_supervision_action_execution.project_id
+    user_id = weak_supervision_action_execution.user_id
+
+    daemon.run(
+        weak_supervision_manager.run_weak_supervision,
```
JWittmeyer, line 90: why are some prestored as variables and other aren'T

JWittmeyer:
Not sure if I saw it right from a time perspective but for the etl Azure DI it seems the queue request is after the extraction. I would assume that the whole process is queued as the extraction also takes up resources (at least for the pdf2md container which apprears to work similar though because it faster i'm not sure. Maybe another point for the ETL bugs for Anmar in the upcoming sprint

- [x] checked
- [x] ignored
- [ ] resolved

JWittmeyer:
Tokenizing seems to be the new queued state for pdf2md or we have to many slots in the queue.
For Azure DI the files are listed as queued, the PUT on the task master however comes once the extraction starts?
Probably related to the point above.

image


- [x] resolved

@JWittmeyer I think there are some misunderstandings how the etl pipeline is working. What I did not change is that the task queue is here only involved at the stage of parsing the file, which we now coined "transformation" if we define the etl pipeline as two steps (1. extraction + 2. transformation). The transformation resembles the states from tokenizing onwards. The extraction is not part of the task queue at all in the current implementation. The state "QUEUE" is kind of misleading, the file is only in this state until the extraction is starting (for pdf2markdown this is almost instanous, but the update is fetched later ui-wise). Then it goes into "TOKENIZATION"-state right before it is put in to the queue.

JWittmeyer:
Heuristics page doesn't work:
image

I'm assuming because we have a join for the table integrated in the request to get the queued state variation:
image
image

Similar for embeddings (and maybe tokenization & attribute calc?), they also checked if it's in queued to prioritize that information over the actual state (NOT_STARTED or similar). Maybe we check for the table name to ensure all queres are changed to the new db shema. I dont think we should invest to much time in the refinery-ui, backend queries should be updated imo

- [x] resolved

@JWittmeyer Did you have an concrete example/error for the embeddings? I checked and found only this example of information source, other usually use the sqllalchemy syntax here

JWittmeyer:
Attribute Calc doesn't work with the queue or something else is wrong. I added 2 (priority slots are 1) and it resulted in an error 
image

Maybe a race condition or something else is up

- [x] resolved

JWittmeyer:
Deleteing a project during the wizard resulted in an error in the task master in the send_org_update. I think we can check for project_item & simply do nothing if it doesn't exist as these are just the websocket notifications 
image

I would have expected the delete query call maybe, though might not be automatic if it's via a cognition project delete with corresponding linked project delete


- [x] resolved

andhreljaKern (approved):

PR #254, opened 2024-09-16
Removal of crowd labeler and zeroshot

lumburovskalina:
Main PR: https://github.com/code-kern-ai/refinery-ui/pull/42


controller/project/manager.py
```diff
@@ -274,11 +274,7 @@ def __no_huddle_id(data_id: str) -> bool:
 
 def __get_crowd_label_is_data(project_id: str, is_id: str) -> Dict[str, str]:
     information_source_item = information_source.get(project_id, is_id)
-    if (
-        not information_source_item
-        or information_source_item.type
-        != enums.InformationSourceType.CROWD_LABELER.value
-    ):
+    if not information_source_item:
```
JWittmeyer, line 277: i think the whole function can be removed

lumburovskalina: yes it can be, 
this is connected with the roles and slices type so i though to remove it with this point

PR #257, opened 2024-10-08
Cache Files for ETL Processing

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-ui/pull/98


api/transfer.py
```diff
@@ -55,7 +55,7 @@ async def post(self, request) -> PlainTextResponse:
 
         parts = file_path.split("/")
 
-        if parts[1] == "_cognition":
+        if parts[1] == "_cognition" and parts[2] == "files":
```
JWittmeyer: so if the second part isn't "files" but the first part still "_cognition" it wont use the cognition route?

Could be right, feels off maybe `or` was meant?

LennartSchmidtKern: What is it doing with congition routes besides of file caching?

JWittmeyer: i dont know, I just saw the code

JWittmeyer (approved):

PR #270, opened 2024-11-18
Users search improvments

lumburovskalina:
Main PR: https://github.com/code-kern-ai/admin-dashboard/pull/63


controller/user/manager.py
```diff
@@ -20,6 +20,7 @@ def get_or_create_user(user_id: str) -> User:
     if not user_item:
         user_item = user.create(user_id, with_commit=True)
         kratos.__refresh_identity_cache()
+        migrate_kratos_users()
```
JWittmeyer: the update of our db is only called if a user doesn't exist yet. This means on startup & after a new user we get the update potentially resulting in an issue for updates on existing users (e.g. verfied status) I think adding it ot eh __refresh_identity_cache function itself will solve this since it's called on a timer

controller/user/manager.py
```diff
@@ -104,53 +105,50 @@ def update_last_interaction(user_id: str) -> None:
...
+
+
+def migrate_kratos_users() -> None:
+    # this is only supposed to be called during startup of the application
+    daemon.run_with_db_token(__migrate_kratos_users)
+
+
+def __migrate_kratos_users():
+    users = kratos.get_cached_values()
+    filtered_users = {k: v for k, v in users.items() if k != "collected"}
```
JWittmeyer: this is only one key. since we also just use it as a lookup this shouldn't be necessary to create (at least not if we simply collect all users from db not by id list)

controller/user/manager.py
```diff
@@ -104,53 +105,50 @@ def update_last_interaction(user_id: str) -> None:
...
+    users_database = user.get_by_id_list(list(filtered_users.keys()))
```
JWittmeyer: why not simply all? id list will be more expensive to plan

controller/user/manager.py
```diff
@@ -104,53 +105,50 @@ def update_last_interaction(user_id: str) -> None:
...
+            != user_identity["verifiable_addresses"][0]["created_at"]
+            or user_database.metadata_public != user_identity["metadata_public"]
+            or (
+                user_identity["metadata_public"]
+                .get("registration_scope", {})
+                .get("provider_id", None)
+                != user_database.sso_provider
+            )
+        ):
+            user.update_user(
```
JWittmeyer: this will update all fields of the user if one is changed. with an user object, setting of that object and commit this wouldn't be the case.

something like:
```python
for user_database in users_database:
    if user_database.email != user_identity["traits"]["email"]:
        user_database.email = user_identity["traits"]["email"]
    if user_database.verfied != user_identity["verifiable_addresses"][0]["verified"]:
       user_database.verfied = user_identity["verifiable_addresses"][0]["verified"]

...
# end of for

general.commit()
```

controller/user/manager.py
```diff
@@ -104,53 +105,50 @@ def update_last_interaction(user_id: str) -> None:
...
+                created_at=user_identity["verifiable_addresses"][0]["created_at"],
+                metadata_public=user_identity["metadata_public"],
+                sso_provider=(
+                    user_identity["metadata_public"]
+                    .get("registration_scope", {})
+                    .get("provider_id", None)
+                    if user_identity["metadata_public"]
+                    else None
+                ),
+                with_commit=True,
```
JWittmeyer: commit on every change is quite expensive usually we do this after a loop (see comment above)

controller/user/manager.py
lumburovskalina: The migrate_kratos_users() called the get_cached_users() which calls __refresh_identity_cache and if we add in in it it comes to a loop, that's why i added it here

fast_api/routes/organization.py
```diff
@@ -281,23 +283,37 @@ def get_mapped_sorted_paginated_users(
...
     ]
-    active_users = {user["id"]: user for user in active_users}
 
-    data, final_len = user_manager.get_mapped_sorted_paginated_users(
-        active_users, body.sort_key, body.sort_direction, body.offset, body.limit
+    final_users = []
+    final_users = sorted(
+        active_users,
+        key=lambda x: (x[body.sort_key] is None, x.get(body.sort_key, "")),
+        reverse=body.sort_direction == -1,
```
JWittmeyer: the sort & pagination should be done on the database for the fields that exist on the database. Less data we need to extend etc. could also already provide the org name so we dont need to worry about it here. 

I can help with that :)

controller/user/manager.py
JWittmeyer: then we need to change the structure a bit imo.

JWittmeyer (approved):

PR #280, opened 2024-12-10
Python SDK Removal

anmarhindi:
#### PRs
- https://github.com/code-kern-ai/refinery-gateway/pull/280
- https://github.com/code-kern-ai/cognition-gateway/pull/184
- https://github.com/code-kern-ai/dev-setup/pull/52
- https://github.com/code-kern-ai/deployment-cognition/pull/7
- https://github.com/code-kern-ai/deployment-managed-cognition/pull/38

#### Post PR approval/merge TODO:
- [ ]  Archive [refinery-gateway-proxy](https://github.com/code-kern-ai/refinery-gateway-proxy) (admin)
- [ ] <del>remove refinery gateway proxy from app_version</del> (can't be found)

#### Start with
`bash start -x -b sdk-removal`


app.py
```diff
@@ -124,13 +120,6 @@
     Route("/notify/{path:path}", Notify),
     Route("/healthcheck", Healthcheck),
     Route("/project/{project_id:str}", ProjectDetails),
```
JWittmeyer: i think some other routes are also not in use anymore. e.g. /import/task/ with UploadTaskInfo was afaik only used in the proxy. Some other might also be irrelevant by now but never were removed. ProjectDetails, CognitionImport & CogntiionPrepareProject since we still have some time lets also remove them where possible.

anmarhindi: Sounds good, looking into removing them 👍

app.py
```diff
@@ -121,25 +113,7 @@
 
 routes = [
     Route("/full_config", FullConfigRest),
-    Route("/notify/{path:path}", Notify),
-    Route("/healthcheck", Healthcheck),
```
JWittmeyer: i think these two are still needed 
Notify is the minio endpoint, called from a different container (e.g. for our project file upload)
Heathcheck is called by the better stack.

Refinery project upload also doesn't work atm in branch

app.py
```diff
@@ -121,25 +113,7 @@
...
-    Route("/project/{project_id:str}", ProjectDetails),
-    Route(
-        "/project/{project_id:str}/knowledge_base/{knowledge_base_id:str}",
-        KnowledgeBaseExport,
-    ),
-    Route("/project/{project_id:str}/associations", AssociationsImport),
-    Route("/project/{project_id:str}/export", FileExport),
-    Route("/project/{project_id:str}/import_file", PrepareFileImport),
     Route("/project/{project_id:str}/import_json", JSONImport),
```
JWittmeyer: we used it for some import from e.g. gates and there is a function in cogntion-gateway that gets removed with this pr group. the refinery wizard also uses it but the wizard isn't in use anymore so i think this can be removed as well.

JWittmeyer: wizard files can also be removed imo

JWittmeyer: probably all files but minio_upload.py in controller->transfer 🤔

app.py
anmarhindi: Nicely spotted, adding those back and retesting with import.

app.py
anmarhindi: Files removed. Apart from minio_upload.py, import_preparator.py is the other file that was kept in controller->transfer->cognition

JWittmeyer (approved):

PR #283, opened 2025-01-02
Forces the datatype of a column after an upload

JWittmeyer:
Forces the datatype of an existing attribute if new data is uploaded. if it can't be forced it fails.

Example files:
1.  numbers file
[dummy_numbers.csv](https://github.com/user-attachments/files/18289961/dummy_numbers.csv)
2. file to extend which didn't work before
[dummy_numbers_extension.csv](https://github.com/user-attachments/files/18289963/dummy_numbers_extension.csv)
3. error case
[dummy_numbers_extension - error.csv](https://github.com/user-attachments/files/18289962/dummy_numbers_extension.-.error.csv)

### DIFF
--- a/controller/transfer/record_transfer_manager.py
+++ b/controller/transfer/record_transfer_manager.py
@@ -71,8 +71,23 @@ def import_records_and_rlas(
             attribute_lookup = {
                 k: v for k, v in attribute_lookup.items() if v is not None
             }
-        force_data_type_for_attributes(records_data, attribute_lookup)
+        try:
+            force_data_type_for_attributes(records_data, attribute_lookup)
+        except Exception as e:
+            logger.error(f"Error while forcing data type for attributes: {e}")
+            if upload_task is not None:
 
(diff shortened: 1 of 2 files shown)


controller/transfer/record_transfer_manager.py
```diff
@@ -81,6 +107,38 @@ def import_records_and_rlas(
             )
 
 
+def force_data_type_for_attributes(
```
lumburovskalina: is there a specific reason why this function is public and the one below is private when both are used in the same file?

lumburovskalina:
I am not getting an error case in the last csv file, i tried to upload it and reupload it as more records.  Or I am doing something wrong?

- [x] ignored
- [ ] resolved

controller/transfer/record_transfer_manager.py
JWittmeyer: no reason => changed to __

lumburovskalina (approved):

PR #287, opened 2025-01-13
LLM attribute calculation

andhreljaKern:
This is the Main PR.

Related:

- https://github.com/code-kern-ai/refinery-submodule-model/pull/151
- https://github.com/code-kern-ai/refinery-gateway/pull/287
- https://github.com/code-kern-ai/refinery-ac-exec-env/pull/70
- https://github.com/code-kern-ai/submodule-javascript-functions/pull/14
- https://github.com/code-kern-ai/refinery-ui/pull/61
- https://github.com/code-kern-ai/refinery-tokenizer/pull/83
- https://github.com/code-kern-ai/submodule-react-components/pull/37


alembic/versions/10c48793371d_adds_attribute_additional_config_and_.py
```diff
@@ -0,0 +1,28 @@
+"""adds attribute `additional_config` and llm_config
```
andhreljaKern: rename file

controller/attribute/manager.py
```diff
@@ -25,6 +25,24 @@
 from . import util
 from sqlalchemy import sql
 
+DEFAULT_LLM_RESPONSE_CONFIG = {
+    "llmIdentifier": "Open AI",
+    "templatePrompt": "You are a chat bot trying to assist a user with a question.",
+    "questionPrompt": "{{ rd.record.question }}",
```
andhreljaKern: ensure it matches usable mustache and parse clientType from llmIdentifier instead of llmConfig.clientType

controller/attribute/util.py
```diff
@@ -9,6 +9,7 @@
 import datetime
 from dateutil import parser
 
+from exceptions.exceptions import LlmConfigError
```
andhreljaKern: Discussion topic: run test_client_config via refinery-gateway route or refinery-ac-exec-env?
@JWittmeyer @DerKernigeFeuerpfeil

controller/attribute/util.py
```diff
@@ -52,7 +55,9 @@ def add_log_to_attribute_logs(
         general.commit()
 
 
-def prepare_sample_records_doc_bin(attribute_id: str, project_id: str) -> str:
+def prepare_sample_records_doc_bin(
+    attribute_id: str, project_id: str
+) -> str:  # TODO: add parameter `n` to limit LLM response executions
```
andhreljaKern: remove comment

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,270 @@
...
+    ## if we reach this point, we couldn't access the value
+    if raise_me:
+        raise ValueError("Couldn't access value from", open_ai_obj)
+    return ""
+
+
+# all work similar but use different classes etc.
+# note that kwargs is just passed to the openai client so adding unknown kwargs will result in issues
+# named parameter are NOT considered kwargs, only unknown parameters are kwargs
+def get_chat_completion(
```
andhreljaKern: Rename functions into less descriptive names (name+UUID) to avoid users calling these by accident

controller/attribute/util.py
```diff
@@ -71,6 +76,42 @@ def prepare_sample_records_doc_bin(attribute_id: str, project_id: str) -> str:
     return prefixed_doc_bin
 
 
+def prepare_llm_response_code(attribute_item: Attribute) -> str:
+    global LLM_RESPONSE_TMPL_PATH
+    with open(LLM_RESPONSE_TMPL_PATH, "r") as file:
+        lines = [line.rstrip() for line in file if line[0] != "#"]
+
+    llm_code = "\n".join(lines)
+    llm_config = attribute_item.additional_config.get("llmConfig", {})
```
andhreljaKern: ensure not-null

controller/attribute/util.py
```diff
@@ -71,6 +74,45 @@ def prepare_sample_records_doc_bin(attribute_id: str, project_id: str) -> str:
...
+        lines = [line.rstrip() for line in file if line[0] != "#"]
+
+    llm_code = "\n".join(lines)
+    if not attribute_item.additional_config:
+        llm_config = {}
+    else:
+        llm_config = attribute_item.additional_config.get("llmConfig", {})
+
+    try:
+        llm_config_mapping = {
```
andhreljaKern, line 246: @DerKernigeFeuerpfeil more explicit error message?

DerKernigeFeuerpfeil: done

controller/attribute/manager.py
```diff
@@ -25,6 +25,24 @@
...
+    "llmConfig": {
+        "model": "gpt-4o-mini",
+        "temperature": 0,
+        "maxLength": 1024,
+        "stopSequences": [],
+        "topP": 1,
+        "frequencyPenalty": 0,
+        "presencePenalty": 0,
+        "apiKey": None,
+        "endpoint": None,
```
andhreljaKern: rename to apiBase

controller/attribute/manager.py
```diff
@@ -128,7 +152,10 @@ def update_attribute(
 def delete_attribute(project_id: str, attribute_id: str) -> None:
```
andhreljaKern, line 155: @JWittmeyer when an attribute is deleted, it does not cleanup the related S3 storage (like we do in https://github.com/code-kern-ai/refinery-gateway/pull/287/files#diff-df72b95bfa70d5b849e68e279c6c60ebd1b2f3c054d31b578fb744ce06921e3aR406). Should we add the same s3 deletion here?

JWittmeyer: yes we should cleanup when reasonable (attribute set to useable & attribute deletion)

andhreljaKern: when an attribute "is_usable", it's S3 related objects are already deleted. The issue we're seeing is deleting an attribute which is not set to useable -> at this point, the S3 related objects persist

JWittmeyer: then deletion it is :)

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+        "api_base": API_BASE_A2VYBG,
+        "api_version": API_VERSION_A2VYBG,
+        "model": MODEL_A2VYBG,
+        "system_prompt": SYSTEM_PROMPT_A2VYBG,
+        "user_prompt": USER_PROMPT_A2VYBG,
+        "llm_kwargs": LLM_KWARGS_A2VYBG,
+    }
+
+
+async def get_llm_response(record: dict, cached_records: dict):
```
anmarhindi, line 324: Related functions are not found in use relating to `llm_resonse_tmpl.py`. Is this file used externally by local scripts that are not part of this repo?

anmarhindi: I can see now that this file is read directly instead.

controller/attribute/util.py
```diff
@@ -71,12 +80,222 @@ def prepare_sample_records_doc_bin(attribute_id: str, project_id: str) -> str:
...
+def prepare_llm_response_code(
+    attribute_item: Attribute,
+    llm_playground_config: Union[Dict[str, Any], None] = None,
+    llm_ac_cache_access_link: Union[str, None] = None,
+    llm_ac_cache_file_upload_link: Union[str, None] = None,
+    num_workers: int = 100,
+    max_api_call_retries: int = 5,
+    retry_sleep_seconds: int = 5,
+) -> str:
+    global LLM_RESPONSE_TMPL_PATH
```
anmarhindi: global can be removed here

controller/attribute/util.py
```diff
@@ -71,12 +80,222 @@ def prepare_sample_records_doc_bin(attribute_id: str, project_id: str) -> str:
...
+        llm_config = {}
+    elif llm_playground_config is None:
+        llm_config = dict(
+            attribute_item.additional_config.get("llmConfig", {}),
+            llmIdentifier=attribute_item.additional_config["llmIdentifier"],
+            templatePrompt=attribute_item.additional_config["templatePrompt"],
+            questionPrompt=attribute_item.additional_config["questionPrompt"],
+            llmAcCacheAccessLink=llm_ac_cache_access_link,
+            llmAcCacheFileUploadLink=llm_ac_cache_file_upload_link,
+        )
```
anmarhindi: Suggestion for simpler condition

```suggestion
    if llm_playground_config is None:
        if not attribute_item.additional_config:
            llm_config = {}
        else
            llm_config = dict(
                attribute_item.additional_config.get("llmConfig", {}),
                llmIdentifier=attribute_item.additional_config["llmIdentifier"],
                templatePrompt=attribute_item.additional_config["templatePrompt"],
                questionPrompt=attribute_item.additional_config["questionPrompt"],
                llmAcCacheAccessLink=llm_ac_cache_access_link,
                llmAcCacheFileUploadLink=llm_ac_cache_file_upload_link,
            )
```

controller/attribute/manager.py
andhreljaKern: changes pushed :)

lumburovskalina:
Not sure if part of this PR, but creating the Conversational AI sample project resulted into error

- [ ] resolved
- [x] ignored

_____________________
Edit @JWittmeyer:
couldn't reproduce

lumburovskalina:
Since this is a Note a suggest that we put into italic 

Screenshot 2025-02-04 at 16 00 00


- [x] resolved
- [ ] ignored

lumburovskalina:
If running id field is empty, the button 'Get by running_id' should be disabled 
Screenshot 2025-02-04 at 16 01 34

- [x] resolved

lumburovskalina:
In the LLM playground modal, I would suggest that the 'Use current values for attribute' should be disabled if the 'Test configuartion' is still running (since it takes some time)

- [x] resolved
- [ ] ignored

lumburovskalina:
If you try to change the Dropdown 'Load config template' it doesn't display the correct value

Screenshot 2025-02-04 at 16 50 22

- [ ] resolved 
- [x] intended

lumburovskalina:
Should we display this in `<pre>` tag?(to look more like code)

Screenshot 2025-02-04 at 16 53 02

- [ ] resolved
- [x] ignored

____________________________
edit @JWittmeyer:
Added for view button display but not in the shorthand display

lumburovskalina:
If i create an attribute from different type (ex. Category) and i change the datatype i don't see any of the configuration for the LLM response (only the provider and the api key), and the code is still the same.

We have some logic  here that if it's the initial value, so nothing has changes since the creation, the user can still override it,
otherwise it's still possible/clickable but it doesn't make too much sense because the user will loose the data.

In this case we only need the config fields and the code change.

- [x] resolved

_________________________
Edit @JWittmeyer 
Disabled the option to switch to LLM Type and vice versa since the attribute needs to be created with a different configuration

fast_api/routes/attribute.py
```diff
@@ -34,6 +35,12 @@ def get_attributes(
 ):
     data = manager.get_all_attributes(project_id, state_filter)
     data_dict = sql_alchemy_to_dict(data, column_whitelist=ALL_ATTRIBUTES_WHITELIST)
+    # removes api key from llmConfig to prevent it from showing in the frontend
+    for attr in data_dict:
+        if attr.get("additional_config"):
+            for k, v in attr["additional_config"].items():
+                if k == "llmConfig":
+                    del v["apiKey"]
```
LennartSchmidtKern: maybe simplify in something like this (not tested)

```
for attr in data_dict:
    attr.get("additional_config", {}).get("llmConfig", {}).pop("apiKey", None)
```

controller/attribute/util.py
andhreljaKern: do you have specific reasoning to remove it? The reason I would keep it is because the variable is defined in global scope. I know the capitalisation makes it clear that it is a global variable, but I would keep it for the purpose of explicity

fast_api/routes/attribute.py
andhreljaKern: tested & accepted

controller/attribute/util.py
andhreljaKern: btw: I am not as familiar with the rest of the codebase, if our convention is to omit "global" for capitalised variable names, that is a good reason IMO

lumburovskalina:
i think the modal can be a bit wider 

Screenshot 2025-02-10 at 16 37 17

- [x] resolved
- [ ] ignored

lumburovskalina:
Test configuration and 'Load config template' still possible for usable attributes, i think it should be disabled. 

Edit Lina: I noticed that most of the fields are enabled, is that wanted?

- [x] resolved
- [ ] ignored

_____________
Edit Jwittmeyer
Playground removed for useable attributes

lumburovskalina:
Cursor pointer on disabled option in dropdown

Screenshot 2025-02-10 at 16 42 30


- [x] resolved

_______________
Edit JWittmeyer
Tooltip cursor overlays. imo best possibility to set to cursor-not-allowed => results in small overleap on edges 

Alternative could be cursor default but then the disabled would also be default

controller/attribute/util.py
```diff
@@ -71,12 +80,222 @@ def prepare_sample_records_doc_bin(attribute_id: str, project_id: str) -> str:
...
+    response.raise_for_status()
+    return response.json()["choices"][0]["message"]["content"]
+
+
+def validate_user_prompt(project_id: str, user_prompt: str):
+    def parse_mustache_attribute_names(mustache_str: str) -> List[str]:
+        for brace in "{}":
+            mustache_str = mustache_str.replace(brace, "")
+        return mustache_str.strip()
+
```
anmarhindi, line 145: the nested function `parse_mustache_attribute_names` can be lifted

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+    global SYSTEM_PROMPT_A2VYBG, USER_PROMPT_A2VYBG, MODEL_A2VYBG, API_KEY_A2VYBG, API_BASE_A2VYBG, API_VERSION_A2VYBG
```
anmarhindi: The global declarations here are not necessary. Line can be removed.

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+        **kwargs,
+    )
+    if close_after:
+        await client.close()
+
+    return completion
+
+
+def get_llm_config_a2vybg():
+    global API_KEY_A2VYBG, API_BASE_A2VYBG, API_VERSION_A2VYBG, CLIENT_TYPE_A2VYBG, MODEL_A2VYBG, SYSTEM_PROMPT_A2VYBG, USER_PROMPT_A2VYBG, LLM_KWARGS_A2VYBG
```
anmarhindi: The global declarations here are not necessary. Should be removed.

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+
+def get_client_8e8a360e_3f7f_4cf9_ba80_8cb239e897d2(
+    use_async: bool,
+    api_key: str,
+    azure_endpoint: Optional[str] = None,
+    api_version: Optional[str] = None,
+    check_valid: bool = True,
+    prevent_cached_client: bool = True,
+) -> Union[OpenAI, AsyncOpenAI, AzureOpenAI, AsyncAzureOpenAI]:
+    global CLIENT_TYPE_A2VYBG
```
anmarhindi: Global declaration not needed here.

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+
+    if CLIENT_TYPE_A2VYBG == LLMProvider_A2VYBG.AZURE.value and (
+        azure_endpoint is None or api_version is None
+    ):
+        raise ValueError("azure_endpoint and api_version must be set for Azure OpenAI")
+
+    # tuples can be used as dict keys, primitive datatype comparison works flawless, caution with objects though!
+    config = (CLIENT_TYPE_A2VYBG, use_async, api_key, azure_endpoint, api_version)
+    use_cache = MAX_CACHED_CLIENTS_A2VYBG != 0 and not prevent_cached_client
+    global CLIENT_LOOKUP_A2VYBG
```
anmarhindi: Global declaration not needed here.

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+        return client
+
+
+def __create_client_bf47529a_75f7_498b_a091_4e7d52d35b6b(
+    use_async: bool,
+    api_key: str,
+    azure_endpoint: Optional[str] = None,
+    api_version: Optional[str] = None,
+):
+    global CLIENT_TYPE_A2VYBG
```
anmarhindi: Global declaration not needed here.

controller/attribute/llm_response_tmpl.py
```diff
@@ -0,0 +1,380 @@
...
+    if not open_ai_obj:
+        return ""
+    if hasattr(open_ai_obj, "choices") and len(open_ai_obj.choices) > 0:
+        t = open_ai_obj.choices[0]
+        if hasattr(t, "message") and hasattr(t.message, "content"):
+            # fmt:off
+            content = t.message.content or '{\"result\": \"Error: N/A\"}'
+            # fmt:on
+            try:
+                content = json.loads(content)
```
anmarhindi, line 222: Perhaps expand the try except to catch possible IndexError for choices

controller/attribute/util.py
andhreljaKern: it is only ever used for validation purposes by the `validate_user_prompt` function, I left is as a nested function to reduce the amount of "global" functions in attribute/util.py. If this is not justified reasoning, i.e. other similar implementations in our codebase lift these kinds of functions, I can lift it

controller/attribute/llm_response_tmpl.py
andhreljaKern: this function ensures an IndexError will not happen with `if hasattr(open_ai_obj, "choices") and len(open_ai_obj.choices) > 0`, i.e. a ValueError is raised with an "Unknown open_ai_obj" message

controller/attribute/llm_response_tmpl.py
andhreljaKern: The get_client function modifies CLIENT_LOOKUP_A2VYBG in a couple `if-else` scenarios so I will keep the global declaration here (but move it to the top of the function body) instead of declaring it in each `if-else` branch

controller/attribute/llm_response_tmpl.py
anmarhindi: Nice, indeed. I must have not highlighted that line too. resolved

controller/attribute/util.py
anmarhindi: If there is no plan to expand and/or use it elsewhere than it can stay.

controller/attribute/llm_response_tmpl.py
anmarhindi: Resolved 👍

JWittmeyer (approved):

PR #288, opened 2025-01-14
Question Playground

anmarhindi:
#### PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/288
- https://github.com/code-kern-ai/refinery-submodule-model/pull/152
- https://github.com/code-kern-ai/refinery-embedder/pull/137
- https://github.com/code-kern-ai/refinery-neural-search/pull/78
- https://github.com/code-kern-ai/refinery-ui/pull/59
- https://github.com/code-kern-ai/submodule-react-components/pull/35
- https://github.com/code-kern-ai/cognition-gateway/pull/191

#### Start with
`./start -x -b question-playground`

#### Example test cases:
* Test different embeddings (Azure etc)
* Test different projects
* Test with different query sizes (multiple sentences in query)
* Test runs with multiple groups

> [!NOTE]  
> For sample projects in Refinery, you must delete and re-create the default embedding. You only need to do this once.


DerKernigeFeuerpfeil:
A - Improvement ideas (dont have to be included now, just for future reference):
1. have card area individually scrollable in playground left and right independent, just like in eval set creation
- [x] noted
2. create eval set has "clear" button for sim search but not for regular search, which I think it should have
- [x] noted
3. inconsistent height in eval set depending on whether I use sim search or regular search (feels jumpy)
- [x] noted

B - Bugs?
1. playground returns records for query X, but query X does not return ANY records in the eval set creation (threshold criterion not met?) - tested 3 embeddings with 3 queries each
- [ ] resolved
- [x] wontfix

Lennart: We did not add a threshold value for the set creation and it will be some default value, while the playground has on default a -9999 (no threshold) value.

2. (probably unrelated to your changes) Screenshot 1 shows the clientside error I get when I try to add an embeddding while another embedding is currently being created
- [ ] resolved
- [x] wontfix

Lennart: I can not reproduce this, I can create embedding and the other one is normally queued. Based on the error message, it seems like some models could not be accessed from catch maybe.

image

alembic/versions/0c8eb3ff1c71_adds_playground_question.py
```diff
@@ -0,0 +1,47 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('playground_question',
+    sa.Column('id', postgresql.UUID(as_uuid=True), nullable=False),
+    sa.Column('question', sa.String(), nullable=True),
+    sa.Column('created_at', sa.DateTime(), nullable=True),
+    sa.Column('created_by', postgresql.UUID(as_uuid=True), nullable=True),
+    sa.Column('project_id', postgresql.UUID(as_uuid=True), nullable=True),
+    sa.Column('embedding_id', postgresql.UUID(as_uuid=True), nullable=True),
+    sa.Column('record_ids', sa.JSON(), nullable=True),
+    sa.Column('meta_info', sa.JSON(), nullable=True),
```
JWittmeyer: not sure i understand the table correct. 
Itended was a table that holds asked questions so they can be rerun against different embeddings and tested so Solution Architects have a set of asked questions to evaluate

anmarhindi: We added the records for the case that in the future its possible to extend by showing the results of previously asked questions, and then selecting a new embedding to get results and do a comparison.

These can be removed to keep it simple and light.

JWittmeyer: i like the idea. we might need to look out for size of db but it shouldn't be to big i hope since we clean up afterwards. => comment with intendet future use would do the trick for me

controller/playground/manager.py
```diff
@@ -0,0 +1,349 @@
...
+from .reformulation import reformulate_question
+import json
+
+NEURAL_SEARCH = os.getenv("NEURAL_SEARCH")
+EMBEDDING_SERVICE = os.getenv("EMBEDDING_SERVICE")
+
+EVALUATION_RUN_LIMIT_DEFAULT = 100
+
+
+class EVALUATION_RUN_STATE:
```
JWittmeyer: maybe to enums in submodule

alembic/versions/0c8eb3ff1c71_adds_playground_question.py
anmarhindi: Ok sounds good. I will comment out these fields for now (will include a comment explaining) and in the future we can enable these at anytime if need be.

controller/playground/manager.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+def get_evaluation_runs(project_id: str):
+    evaluation_runs_objects = evaluation_run_db_bo.get_all(project_id)
+    evaluation_runs = []
+    for evaluation_run in evaluation_runs_objects:
+        results = to_frontend_obj_raw(evaluation_run.results)
+        evaluation_runs.append(
+            {**sql_alchemy_to_dict(evaluation_run, True), "results": results}
+        )
```
JWittmeyer: dublicated code (get result to dict) maybe as small function + list comprehension

controller/playground/manager.py
```diff
@@ -0,0 +1,349 @@
...
+    return evaluation_runs
+
+
+def get_evaluation_run_by_id(project_id: str, run_id: str):
+    evaluation_run = evaluation_run_db_bo.get(project_id, run_id)
+    results = to_frontend_obj_raw(evaluation_run.results)
+    evaluation_run_object = {
+        **sql_alchemy_to_dict(evaluation_run, True),
+        "results": results,
+    }
```
JWittmeyer: if function from above is created here as well

controller/playground/manager.py
```diff
@@ -0,0 +1,349 @@
...
+        headers={
+            "accept": "application/json",
+            "content-type": "application/json",
+        },
+        json={"texts": texts},
+    )
+    if response.ok:
+        return response.json()["tensor"]
+    else:
+        return None
```
JWittmeyer: usually we use service_requests.post_call_or_raise in a new function inthe controller>embedding>connector.py for requests between our containers

controller/playground/manager.py
```diff
@@ -0,0 +1,349 @@
...
+    embedding_id: str,
+    embedding_tensor: List[float],
+    limit: int,
+    similarity_filter_option: Optional[List[Dict[str, Any]]] = None,
+    threshold: Optional[float] = None,
+    save_question: Optional[bool] = None,
+    question: Optional[str] = None,
+    user_id: Optional[str] = None,
+):
+    url = f"{NEURAL_SEARCH}/most_similar_by_embedding?include_scores=true"
```
JWittmeyer: see above

fast_api/routes/playground.py
```diff
@@ -0,0 +1,274 @@
...
+    user_id = auth_manager.get_user_id_by_info(request.state.info)
+    search_results = playground_manager.get_search_result_for_text(
+        project_id,
+        str(user_id),
+        search_question.embeddingId,
+        search_question.question,
+        search_question.limit,
+        search_question.filter,
+        search_question.threshold,
+        search_question.saveQuestion,
```
JWittmeyer: should default to true imo

fast_api/routes/playground.py
```diff
@@ -0,0 +1,274 @@
...
+        evaluation_group.name,
+        evaluation_group.evaluationSetIds,
+        user_id,
+    )
+    return get_silent_success()
+
+
+@router.delete(
+    "/{project_id}/evaluation-groups"
+)  # dependencies=[Depends(auth_manager.check_project_access_dep)]
```
JWittmeyer: comment

JWittmeyer: also probably needed as the others all use the check

fast_api/routes/project_setting.py
```diff
@@ -148,6 +149,27 @@ def get_record_by_record_id(
...
+    "/{project_id}/records-batch",
+    dependencies=[Depends(auth_manager.check_project_access_dep)],
+)
+def get_records_batch(
+    project_id: str,
+    recordsBatchBody: RecordsBatchBody = Body(...),
+):
+    results = []
+    records = record_manager.get_record_by_ids(project_id, recordsBatchBody.record_ids)
+    for record in records:
```
JWittmeyer: list comprehension

controller/playground/manager.py
anmarhindi: Interesting, looks like I already created this class in enums submodule weeks ago but was not used. will adjust

lumburovskalina (approved):

PR #300, opened 2025-04-24
Added full admin access table and endpoint

lumburovskalina:
Main PR: https://github.com/code-kern-ai/admin-dashboard/pull/75


controller/auth/kratos.py
```diff
@@ -191,3 +197,42 @@ def resolve_user_name_and_email_by_id(user_id: str) -> dict:
...
+    response_create = requests.post(
+        f"{KRATOS_ADMIN_URL}/identities",
+        json=payload_registration,
+    )
+    return response_create.json() if response_create.ok else None
+
+
+def get_recovery_link(user_id: str) -> str:
+    payload_recovery_link = {
+        "expires_in": "24h",
```
JWittmeyer: as discussed 48h

controller/auth/manager.py
```diff
@@ -152,3 +154,34 @@ def extract_state_info(request: Request, key: str) -> Any:
...
+    if check_is_admin(request):
+        jwt_decoded: Dict[str, Any] = jwt.decode(
+            request.headers["Authorization"].split(" ")[1],
+            options={"verify_signature": False},
+        )
+        subject: Dict[str, Any] = jwt_decoded["session"]["identity"]
+        if check_email_in_full_admin(subject["traits"]["email"]):
+            return True
+    return False
+
```
JWittmeyer, line 174: can we add a check_invite_users function? I'm thinking of ensureing the email addresses are following a common email regex & dont have an account yet in kratos so if someone pastes an already existing we dont send. On the frontent the button for inviting should only be enabled if the check was successful

fast_api/models.py
```diff
@@ -501,3 +501,8 @@ class EvaluationRunDeletionBody(BaseModel):
 class SearchQuestionReformulationBody(BaseModel):
     question: StrictStr
     apiKey: StrictStr
+
+
+class InviteUsersBody(BaseModel):
+    emails: List[StrictStr]
+    organization_name: StrictStr
```
JWittmeyer, line 508: maybe organization_id? Should reduce the requests we need to do for it to work :)

fast_api/routes/misc.py
```diff
@@ -271,3 +272,16 @@ def update_customer_buttons(
...
+
+@router.get("/is-full-admin")
+def get_is_admin(request: Request) -> Dict:
+    data = auth.check_is_full_admin(request)
+    return pack_json_result(data)
+
+
+@router.post("/invite-users")
+def invite_users(request: Request, body: InviteUsersBody = Body(...)):
+    auth.check_admin_access(request.state.info)
```
JWittmeyer: should be full admin otherwise a none full admin could use it as well

controller/auth/manager.py
lumburovskalina: yes sure, i will add this

fast_api/models.py
lumburovskalina: we use 'update_organization_of_user' with a organization_name as a parameter to update the organization of a user (dropdown on users page), that's why i used the name of the org, we can of course change it?

JWittmeyer: works for me, either way is fine, in theory a name based selection will take longer but since we have max 100 orgs it's a micro optimization so can be skipped :)

controller/auth/manager.py
lumburovskalina: should this be in its own endpoint, that is called on 'Continue' from the modal ?

JWittmeyer: sounds good :)

controller/auth/kratos.py
```diff
@@ -191,3 +197,54 @@ def resolve_user_name_and_email_by_id(user_id: str) -> dict:
...
+    )
+    return response_link.json() if response_link.ok else None
+
+
+def email_with_link(to_email: str, recovery_link: str) -> None:
+    msg = MIMEText(
+        f"Welcome! Click the link to complete your account setup:\n\n{recovery_link}"
+    )
+    msg["Subject"] = "You're invited to our app!"
+    msg["From"] = "no-reply@kern.com"
```
andhreljaKern: should we use `no-reply@kern.ai` instead of `kern.com`?

lumburovskalina: changed

controller/auth/manager.py
```diff
@@ -152,3 +155,52 @@ def extract_state_info(request: Request, key: str) -> Any:
...
+
+def invite_users(emails: List[str], organization_name: str):
+    if not check_is_full_admin:
```
andhreljaKern: The if statement doesn't call the `check_is_full_admin` function, it just checks if the function exists

controller/auth/manager.py
```diff
@@ -152,3 +155,52 @@ def extract_state_info(request: Request, key: str) -> Any:
...
+        recovery_link = kratos.get_recovery_link(user["id"])
+        if not recovery_link:
+            raise AuthManagerError("Failed to get recovery link")
+
+        # Send the recovery link to the email
+        kratos.email_with_link(email, recovery_link["recovery_link"])
+
+
+def check_valid_emails(emails: List[str]):
+    if not check_is_full_admin:
```
andhreljaKern: same as line #173

controller/auth/manager.py
```diff
@@ -152,3 +155,51 @@ def extract_state_info(request: Request, key: str) -> Any:
...
+    for email in emails:
+        if not is_valid_email(email):
+            return False
+        if kratos.check_user_exists(email):
+            return False
+    return True
+
+
+def is_valid_email(email):
+    pattern = r"^[\w\.-]+@[\w\.-]+\.\w+$"
```
JWittmeyer: this creates the pattern on every call of is_valid_email.

Since the regex doesn't change it's better to pre compile it outside the function scope & use fullmatch so it's not jsut a part of the string something like:

```
import re

EMAIL_RE = re.compile(r"[\w\.-]+@[\w\.-]+\.\w+")

def is_valid_email(email: str) -> bool:
    return bool(EMAIL_RE.fullmatch(email))
```

controller/auth/manager.py
```diff
@@ -152,3 +155,51 @@ def extract_state_info(request: Request, key: str) -> Any:
...
+
+    for email in emails:
+        if not is_valid_email(email):
+            return False
+        if kratos.check_user_exists(email):
+            return False
+    return True
+
+
+def is_valid_email(email):
```
JWittmeyer: typing

fast_api/routes/misc.py
```diff
@@ -271,3 +274,25 @@ def update_customer_buttons(
             update_request.visible,
         )
     )
+
+
+@router.get("/is-full-admin")
+def get_is_admin(request: Request) -> Dict:
```
andhreljaKern: Function name already exists

JWittmeyer (approved):

PR #310, opened 2025-06-23
Cognition UI Facelift

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-ui/pull/146


alembic/versions/ad13850a7245_ui_flip_project_icon.py
```diff
@@ -0,0 +1,43 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "user", sa.Column("use_new_cognition_ui", sa.Boolean(), nullable=True)
+    )
+    op.add_column(
+        "project", sa.Column("icon", sa.String(), nullable=True), schema="cognition"
+    )
+    # ### end Alembic commands ###
+
+    update_dataset_sql = """
```
lumburovskalina, line 31: is this the update for the old projects? (no updater needed for the next release?)

fast_api/routes/organization.py
```diff
@@ -248,9 +248,18 @@ def archive_admin_message(
 
 
 # in use cognition-ui (08.01.25)
-@router.post("/set-language-display")
-def set_language_display(request: Request, body: UserLanguageDisplay = Body(...)):
-    user_manager.update_user_language_display(body.user_id, body.language_display)
+@router.put("/set-language-display/{language}")
+def set_language_display(request: Request, language: str):
+    user_id = auth_manager.get_user_id_by_info(request.state.info)
+    user_manager.update_user_language_display(user_id, language)
```
lumburovskalina: maybe we can rename this with the word `set` instead of `update` since we use it everywhere as set

or merge both `update_user_language_display` and `set_use_new_ui` into one and add key value parameter to update the correct field

alembic/versions/ad13850a7245_ui_flip_project_icon.py
JWittmeyer: yes, no updater needed :)

lumburovskalina (approved):

PR #316, opened 2025-07-07
v1.20.0

JWittmeyer:
main PR: https://github.com/code-kern-ai/refinery-ui/pull/76

### DIFF
--- a/controller/misc/manager.py
+++ b/controller/misc/manager.py
@@ -93,7 +93,7 @@ def finalize_customer_buttons(
     for e in buttons:
         e[key_name] = name_lookup[str(e[key])]
         e[key_name] = (
-            (e[key_name]["first"] or "" + " " + e[key_name]["last"] or "")
+            (e[key_name].get("first", "") + " " + e[key_name].get("last", ""))
             if e[key_name]
             else "Unknown"
         )
--- a/controller/misc/manager.py
+++ b/controller/misc/manager.py
@@ -93,7 +93,7 @@ def finalize_customer_buttons(
     for e in buttons:
         e[key_name] = name_lookup[str(e[key])]
         e[key_name] = (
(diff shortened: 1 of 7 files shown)


controller/misc/manager.py
```diff
@@ -93,7 +93,7 @@ def finalize_customer_buttons(
     for e in buttons:
         e[key_name] = name_lookup[str(e[key])]
         e[key_name] = (
-            (e[key_name]["first"] + " " + e[key_name]["last"])
+            (e[key_name]["first"] or "" + " " + e[key_name]["last"] or "")
```
andhreljaKern: @JWittmeyer - not sure if this is a proper solution. Ref: https://www.notion.so/kern-ai/Test-Findings-refinery-1-20-0-cognition-1-8-0-223007cbb4c080099e3dcc73e4eb971b?source=copy_link#229007cbb4c080a89f0be744db85da66

controller/project/manager.py
```diff
@@ -155,6 +155,11 @@ def get_all_projects_by_user(organization_id) -> List[Project]:
         else:
             p["num_data_scale_uploaded"] = record.get_count_scale_uploaded(p["id"])
 
+        p["is_integration_project"] = any(
```
andhreljaKern: We could do this instead
```python
p["is_integration_project"] = bool(
    integration_db_bo.get_all_by_project_id(p["id"])
)
```

controller/misc/manager.py
JWittmeyer: when can they be empty? also if we want a default value a get("first","") might be more readable

controller/project/manager.py
JWittmeyer: new db request vs in memory 
however names can be updated so probably best to be sure with the db query

lumburovskalina (approved):

PR #337, opened 2025-07-24
Fixes no api key embedding name endpoint

JWittmeyer:
main PR: https://github.com/code-kern-ai/cognition-gateway/pull/239

### DIFF
--- a/controller/attribute/util.py
+++ b/controller/attribute/util.py
@@ -139,6 +139,29 @@ def test_azure_foundry_llm_connection(api_key: str, base_endpoint: str):
     return response.json()["choices"][0]["message"]["content"]
 
 
+def test_privatemode_ai_llm_connection(model: str):
+    # more here: https://docs.privatemode.ai/api/chat-completions
+    headers = {
+        "Content-Type": "application/json",
+        # "Authorization": f"Bearer {api_key}",
+    }
+    payload = {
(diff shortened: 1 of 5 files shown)


controller/organization/manager.py
```diff
@@ -101,13 +101,15 @@ def get_overview_stats(org_id: str) -> List[Dict[str, Union[str, int]]]:
 
 # INFO: Not fully debounced if server runs multiple instances
 # TODO: Change to 60 to 300 for prod
-@param_debounce(seconds=60)
+@param_debounce(seconds=300)
```
andhreljaKern, line 104: should we use an env variable rather than a constant?

andhreljaKern (approved):

controller/attribute/util.py
```diff
@@ -139,6 +139,29 @@ def test_azure_foundry_llm_connection(api_key: str, base_endpoint: str):
     return response.json()["choices"][0]["message"]["content"]
 
 
+def test_privatemode_ai_llm_connection(model: str):
+    # more here: https://docs.privatemode.ai/api/chat-completions
+    headers = {
+        "Content-Type": "application/json",
+        # "Authorization": f"Bearer {api_key}",
```
lumburovskalina: comment

PR #344, opened 2025-09-15
Fix `null` created by

andhreljaKern:
Main PR: https://github.com/code-kern-ai/cognition-integration-provider/pull/40

### DIFF
--- a/alembic/versions/58c0e071db99_adds_deleted_user.py
+++ b/alembic/versions/58c0e071db99_adds_deleted_user.py
@@ -7,7 +7,7 @@ Create Date: 2025-09-15 14:08:25.901703
 """
 
 from alembic import op
-import sqlalchemy as sa
+from submodules.model import DELETED_USER_ID
 
 
 # revision identifiers, used by Alembic.
@@ -15,8 +15,6 @@ revision = "58c0e071db99"
 down_revision = "de396670d10f"
 branch_labels = None
 depends_on = None
-
-DELETED_USER_ID = "00000000-0000-0000-0000-000000000000"
 
 
 def upgrade():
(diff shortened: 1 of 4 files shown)


alembic/versions/58c0e071db99_adds_deleted_user.py
```diff
@@ -0,0 +1,61 @@
...
+	sso_provider,
+	use_new_cognition_ui,
+	oidc_identifier
+) values (
+	'{DELETED_USER_ID}', 
+	NULL, 
+	NULL, 
+	NULL, 
+	NULL, 
+	'deleted.user@domain.ai',
```
andhreljaKern: I didn't want to use @kern.ai domain because it has admin privileges in our app

alembic/versions/58c0e071db99_adds_deleted_user.py
```diff
@@ -0,0 +1,61 @@
...
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "58c0e071db99"
+down_revision = "de396670d10f"
+branch_labels = None
+depends_on = None
+
+DELETED_USER_ID = "00000000-0000-0000-0000-000000000000"
```
JWittmeyer: let's put the user id in the model submodule so all repos have access to it

alembic/versions/58c0e071db99_adds_deleted_user.py
JWittmeyer: domain itself isn't enough, we also need a verified kratos user for this id to have any right so it can be @kern.ai

JWittmeyer (approved):

PR #345, opened 2025-09-16
Hotfix permission batch

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-integration-provider/pull/41


fast_api/routes/record_internal.py
```diff
@@ -32,7 +32,16 @@ def sync_access_groups_and_users_sharepoint(
     record_ids = body.get("record_ids")
     integration_id = body.get("integration_id")
     permissions_user = body.get("permissions_user", {})
-
+    if len(project_id ) != 36:
```
andhreljaKern, line 35: rather try to cast to uuid and catch an exception than check for length

JWittmeyer: i prefer the length check if it's sufficient try: catch with expected fails is pythoning but more expensive performance wise

JWittmeyer (approved):

PR #352, opened 2025-10-01
Prevent kern users from being deleted on org deletion

JWittmeyer:


lumburovskalina:
Not sure if wanted or not, the users are deleted only from our database but not from kratos. Is that wanted?

- [ ] resolved
- [x] ignored

lumburovskalina (approved):

PR #362, opened 2025-11-06
Inbox mail table

lumburovskalina:


controller/auth/kratos.py
```diff
@@ -293,3 +294,49 @@ def check_user_exists(email: str) -> bool:
             if i["traits"]["email"].lower() == email.lower():
                 return True
     return False
+
+
+def migrate_kratos_users() -> None:
+    daemon.run_with_db_token(__migrate_kratos_users)
+
+
+def __migrate_kratos_users():
```
lumburovskalina: why do we have two functions `__migrate_kratos_users` ?
remove the old one if not used?

controller/inbox_mail/manager.py
```diff
@@ -0,0 +1,198 @@
...
+        thread_id=thread_id,
+        is_important=is_important,
+        is_admin_support_thread=is_admin_support_thread,
+    )
+
+
+def get_inbox_mails_by_thread(
+    org_id: str,
+    user_id: str,
+    thread_id: Optional[str] = None,
```
lumburovskalina: optional as the last param

controller/organization/manager.py
```diff
@@ -55,15 +54,23 @@ def get_user_count(organization_id: str) -> int:
 
 
 def get_all_users(
-    organization_id: str, user_role: Optional[str] = None, as_dict: bool = True
+    organization_id: str,
+    user_role: Optional[str] = None,
```
lumburovskalina: optional at the end

controller/user/manager.py
```diff
@@ -37,6 +38,28 @@ def get_user_roles() -> Dict[str, str]:
     return {str(u.id): u.role for u in user.get_all()}
 
 
+def get_admin_users(expand_mail_name: bool = False) -> List[User]:
```
lumburovskalina, line 41: this two functions look similar only the call is different not sure if it makes sense to be one function with a parameter?

fast_api/routes/inbox_mail.py
```diff
@@ -0,0 +1,147 @@
...
+            (str(inbox_mail_thread.organization_id) != str(user.organization_id))
+            or not inbox_mail_go.get_inbox_mail_thread_association_by_thread_id_and_user_id(
+                thread_id=inbox_mail.threadId,
+                user_id=str(user.id),
+            )
+        ):
+            raise HTTPException(status_code=403, detail="Not authorized")
+
+    mail = inbox_mail_manager.create_inbox_mail_by_thread(
+        org_id=user.organization_id,
```
lumburovskalina, line 41: shouldn't the organization come from the FE? I think it's not the org of the user but the org that was selected in the modal?

LennartSchmidtKern: the org in the modal is only for selecting/prefiltering the recipient
The org of the thread is determined by the thread creator. Answering threads is also not using the org, because the mail does not have a org, its only set on the thread level

controller/inbox_mail/manager.py
```diff
@@ -0,0 +1,198 @@
...
+            mail_dict["recipientNames"] = [DEFAULT_SYSTEM_AUTOGENERATED_NAME]
+        return
+    if (user_is_sender and not user_is_admin) or (
+        user_is_admin and not sender_is_admin
+    ):
+        mail_dict["senderName"] = kratos.resolve_user_name_by_id(mail_dict["sender_id"])
+        mail_dict["recipientNames"] = [DEFAULT_KERN_AI_ADMIN_NAME]
+
+    elif not user_is_sender and not user_is_admin:
+        mail_dict["senderName"] = DEFAULT_KERN_AI_ADMIN_NAME
```
JWittmeyer, line 185: sender is admin if user_is_admin is false? inteded?

LennartSchmidtKern: its the recipient perspective of a user receiving a message from kern ai support team.

LennartSchmidtKern (approved):

PR #373, opened 2026-01-19
Webpage Integration

LennartSchmidtKern:
Main PR:
- https://github.com/code-kern-ai/cognition-integration-provider/pull/71


alembic/versions/d454676ee76f_adds_web_integrations.py
```diff
@@ -0,0 +1,157 @@
...
+        ["updated_by"],
+        unique=False,
+        schema="integration",
+    )
+    op.create_index(
+        op.f("ix_integration_sharepoint_object_id"),
+        "sharepoint",
+        ["object_id"],
+        unique=False,
+        schema="integration",
```
JWittmeyer, line 102: 🔵 **NOTE**: The migration file is not present in this workspace, so the constraint and table structure could not be inspected. The concern is logically valid: adding a UNIQUE constraint on (integration_id, running_id, object_id, etl_task_id) will fail on upgrade if duplicates exist. Recommend verifying in the repo that contains this migration—either run a pre-upgrade duplicate check or add a data migration to deduplicate before applying the constraint.

alembic/versions/d454676ee76f_adds_web_integrations.py
```diff
@@ -0,0 +1,157 @@
...
+        sa.Column("running_id", sa.Integer(), nullable=True),
+        sa.Column("source", sa.String(), nullable=True),
+        sa.Column("minio_file_name", sa.String(), nullable=True),
+        sa.Column("error_message", sa.String(), nullable=True),
+        sa.Column("extension", sa.String(), nullable=True),
+        sa.Column("title", sa.String(), nullable=True),
+        sa.Column("raw_markdown_content_hash", sa.String(), nullable=True),
+        sa.Column("etag", sa.String(), nullable=True),
+        sa.Column("last_modified", sa.String(), nullable=True),
+        sa.Column("etl_task_id", postgresql.UUID(as_uuid=True), nullable=True),
```
JWittmeyer, line 39: 🔵 **NOTE**: The migration file is not present in this workspace, so the downgrade implementation could not be inspected. The concern is generally valid: if downgrade drops the webpage table, all webpage data is lost. Recommend adding a docstring or comment in the migration that states downgrade causes data loss, so operators know before reverting in environments with data.

start
```diff
@@ -123,6 +123,7 @@ docker run -d --rm \
 -e SMTP_HOST=mailhog \
 -e SMTP_PORT=1025 \
 -e ENABLE_TELEMETRY=$ENABLE_TELEMETRY \
+-e PYDEVD_DISABLE_FILE_VALIDATION=1 \
```
JWittmeyer, line 126: ❓ **QUESTION**: PYDEVD_DISABLE_FILE_VALIDATION=1 is a PyDev debugger setting. Is this required for a specific local/debug workflow, and should it be conditional (e.g. only when a debug flag is set) or documented so it is not enabled in production by default?

JWittmeyer (commented):
Automated review (review-bot)

alembic/versions/d454676ee76f_adds_web_integrations.py
JWittmeyer, line 1: 🔵 **NOTE**: This revision both creates the new webpage table and modifies the existing sharepoint table (index + unique constraint). Consider documenting both in the revision docstring or splitting into two revisions for clearer history.

JWittmeyer:
🤖 Reviewed as part of PR group anchored at [code-kern-ai/cognition-integration-provider#71](https://github.com/code-kern-ai/cognition-integration-provider/pull/71). See the main PR for the full summary.

JWittmeyer (approved):

PR #375, opened 2026-01-20
Allows selection of etl config

JWittmeyer:
PRs:
- https://github.com/code-kern-ai/refinery-gateway/pull/375
- https://github.com/code-kern-ai/cognition-gateway/pull/298
- https://github.com/code-kern-ai/cognition-ui/pull/194


lumburovskalina:
Is this scenario that can happen? 
-> If yes i would expected that only these two configs are displayed in the chat 
-> If not (scenario that all added configs are visible, but one is preselected? , then i think it's a bug)

Screenshot 2026-01-20 at 16 03 47

- [x] resolved
- [ ] discussed
- [ ] ignored

lumburovskalina (approved):

PR #413, opened 2026-05-26
Docker hardened images

lumburovskalina:
Main PR: https://github.com/code-kern-ai/refinery-ui/pull/104


.drone.yml
```diff
@@ -76,13 +83,20 @@ steps:
       - 'git config --global url."https://github.com/".insteadOf git@github.com:'
       - "git submodule update --recursive"
   - name: build and publish
-    image: plugins/docker
+    image: plugins/docker:21.2.8-linux-amd64
```
andhreljaKern:
```suggestion
    image: plugins/docker:21.2.8-linux-arm64
```

.drone.yml
```diff
@@ -140,12 +161,19 @@ steps:
...
```
andhreljaKern:
```suggestion
    image: plugins/docker:21.2.8-linux-arm64
```

dev.Dockerfile
```diff
@@ -17,4 +21,6 @@ RUN pip3 install --no-cache-dir -r requirements-dev.txt
 
 COPY / .
 
-CMD [ "/usr/local/bin/uvicorn", "--host", "0.0.0.0", "--port", "80", "app:app", "--reload" ]
+USER 65532:65532
```
JWittmeyer: 🔴 **CRITICAL**: Switching the dev container to USER 65532:65532 while leaving `COPY / .` unchanged (files remain root-owned) is likely to break `uvicorn --reload`, which needs to write `__pycache__` and other artifacts under `/app`. Expect permission errors at runtime unless the volume mount or an explicit `--chown` makes `/app` writable by UID 65532.

```suggestion
COPY --chown=65532:65532 . .

USER 65532:65532
```

Dockerfile
```diff
@@ -1,9 +1,20 @@
-FROM kernai/refinery-parent-images:v2.5.0-common
+ARG PARENT_IMAGE=kernai/refinery-parent-images:v2.5.0-common
```
JWittmeyer: 🔵 **NOTE**: The production Dockerfile default remains `PARENT_IMAGE=kernai/refinery-parent-images:v2.5.0-common` and the visible amd64 push pipeline does not pass a `PARENT_IMAGE` build-arg. Instead, `.drone.yml` configures `base_image_registry: registry.dev.kern.ai` with `base_image_username`/`base_image_password` from `docker_username`/`docker_password`, which is the Drone docker plugin mechanism for authenticating and pulling `FROM` base images from the private registry. That answers the mechanism question for the push pipeline shown, but dev.Dockerfile, tag/release pipelines, and proof that the private-registry tag is the hardened v2.5.0-common image were not in the provided evidence.

Dockerfile
```diff
@@ -13,6 +24,20 @@ COPY requirements.txt .
...
+ENV VENV_PATH=/opt/venv
+ENV PATH="${VENV_PATH}/bin:${PATH}"
+
+WORKDIR /app
+
+USER root
+
+COPY --from=builder --chown=65532:65532 ${VENV_PATH} ${VENV_PATH}
+COPY --from=builder --chown=65532:65532 /app /app
+
```
JWittmeyer, line 47: 🔵 **NOTE**: The runtime stage sets `USER 65532:65532` and copies artifacts with `--chown=65532:65532`, but the provided evidence does not include the `refinery-parent-images:v2.5.0-common` parent Dockerfile or any definition of its default user/group. Parent-image settings only show naming conventions (`kernai/refinery-parent-images:{version}-common`). UID/GID alignment with the base image contract cannot be confirmed or refuted from the supplied snippets.

Dockerfile
```diff
@@ -13,6 +24,20 @@ COPY requirements.txt .
 
 RUN pip3 install --no-cache-dir -r requirements.txt
 
-COPY / .
+COPY . .
```
JWittmeyer, line 29: 🔵 **NOTE**: The build pattern is confirmed: the builder runs `COPY . .` and the runtime stage copies the entire builder `/app` tree with `COPY --from=builder --chown=65532:65532 /app /app`, so anything in the build context not excluded would reach the final image. However, `.dockerignore` was not found in the provided cloned-repository evidence, so whether secrets and dev artifacts are actually excluded cannot be verified here.

Dockerfile
```diff
@@ -13,6 +24,20 @@ COPY requirements.txt .
...
+
+FROM ${PARENT_IMAGE}
```
JWittmeyer, line 37: 🔵 **NOTE**: Security improvement: build-only tooling (apt packages, Rust toolchain, gcc) stays in the builder stage and is not copied to runtime. Runtime attack surface is reduced versus the previous single-stage Dockerfile, though the full parent base image filesystem is still present in the final stage.

.drone.yml
```diff
@@ -111,12 +125,19 @@ steps:
...
+    depends_on:
+      - submodules
     settings:
       username:
         from_secret: dockerhub_username
       password:
         from_secret: dockerhub_password
+      base_image_registry: registry.dev.kern.ai
```
JWittmeyer: 🔵 **NOTE**: The visible amd64 push pipeline supplies `docker_username`/`docker_password` for both registry push auth and `base_image_registry` pull auth; no DockerHub publish step using `dockerhub_*` secrets appears in the provided (truncated) `.drone.yml`. Tag/release pipeline contexts called out in the finding were not included in the evidence, so availability of registry credentials for those builds cannot be confirmed or denied.

.drone.yml
```diff
@@ -13,13 +13,20 @@ steps:
       - 'git config --global url."https://github.com/".insteadOf git@github.com:'
       - "git submodule update --recursive"
   - name: build and publish
-    image: plugins/docker
+    image: plugins/docker:21.2.8-linux-amd64
```
JWittmeyer: 🔵 **NOTE**: Pinning `plugins/docker:21.2.8-linux-amd64` and adding explicit `depends_on: submodules` improves reproducibility and ordering. The amd64 plugin tag on arm64 pipelines is normal if buildx/platform is configured elsewhere in the pipeline.

dev.Dockerfile
```diff
@@ -1,11 +1,15 @@
...
+ARG PARENT_IMAGE=registry.dev.kern.ai/code-kern-ai/refinery-parent-images:dev-common
+FROM ${PARENT_IMAGE}
 
 WORKDIR /app
 
 VOLUME ["/app"]
 
-# used for encryption and zipping of files
-RUN apt-get update && apt-get install -y curl libc6-dev zlib1g gcc --no-install-recommends
+USER root
```
JWittmeyer, line 8: 🔵 **NOTE**: Dev image remains single-stage and retains curl/gcc/Rust in the final image. Acceptable for local development, but it is not equivalent to production hardening; avoid using `dev.Dockerfile` as a security reference for deployed images.

Dockerfile
JWittmeyer: 🟡 **SUGGESTION**: No HEALTHCHECK was added to the production image. Per the Docker review policy, define a health check (e.g. HTTP probe on port 80) so orchestrators can detect unhealthy instances without relying solely on process presence.

```
HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
  CMD curl -f http://127.0.0.1:80/health || exit 1
```

JWittmeyer:
🤖 Reviewed as part of PR group anchored at [code-kern-ai/refinery-ui#104](https://github.com/code-kern-ai/refinery-ui/pull/104). See the main PR for the full summary.

JWittmeyer (approved):

.drone.yml
```diff
@@ -13,13 +13,20 @@ steps:
...
+    depends_on:
+      - submodules
     settings:
       registry: registry.dev.kern.ai
       username:
         from_secret: docker_username
       password:
         from_secret: docker_password
+      base_image_registry: registry.dev.kern.ai
```
andhreljaKern:
```suggestion
      base_image_registry: dhi.io
```

.drone.yml
```diff
@@ -76,13 +83,20 @@ steps:
...
+    image: plugins/docker:21.2.8-linux-arm64
+    depends_on:
+      - submodules
     settings:
       registry: registry.dev.kern.ai
       username:
         from_secret: docker_username
       password:
         from_secret: docker_password
+      base_image_registry: registry.dev.kern.ai
```
andhreljaKern:
```suggestion
      base_image_registry: dhi.io
```

.drone.yml
andhreljaKern:
```suggestion
      base_image_registry: dhi.io
```

.drone.yml
```diff
@@ -140,12 +161,19 @@ steps:
...
-    image: plugins/docker
+    image: plugins/docker:21.2.8-linux-arm64
+    depends_on:
+      - submodules
     settings:
       username:
         from_secret: dockerhub_username
       password:
         from_secret: dockerhub_password
+      base_image_registry: registry.dev.kern.ai
```
andhreljaKern:
```suggestion
      base_image_registry: dhi.io
```

